PR #16, opened 2019-09-26
Dockerfile and compose improvements

pawelwilczynski:
Dockerfile and compose improvements

* use_alpine=y option added to build app container based on python:3.6-alpine
* docker image is now multistage - what should shorten its build time
* reaping zombie processes and forwarding signals is now done by docker's
  '--init' option (Tini init) and not 's6' init binary
* newer docker-compose 1.22.0+ is needed (docker engine 18.06.0+)
  Follow instructions from https://docs.docker.com/compose/install/#install-compose


{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,41 @@
...
+
 COPY ./src/ /srv/app/
 
 WORKDIR /srv/app
 
-RUN pip3 install --no-cache-dir -r requirements.txt gunicorn\
-    && python3 -m compileall -b .
+ENV PATH=/root/.local/bin:$PATH
 
-ENTRYPOINT ["/init"]
```
pawelwilczynski: I can't understand what is or was `/init`. I can't find it anywhere in this repo or `python` image itself.

ppolewicz: that looks like a part of s6-overlay

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,41 @@
...
+RUN python3 -m compileall -b .
```
ppolewicz: shouldn't this happen on the first stage?

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,41 @@
...
+{%- endif %}
+COPY --from=base-image /root/.local /root/.local
+
 COPY ./src/ /srv/app/
 
 WORKDIR /srv/app
 
-RUN pip3 install --no-cache-dir -r requirements.txt gunicorn\
-    && python3 -m compileall -b .
+ENV PATH=/root/.local/bin:$PATH
```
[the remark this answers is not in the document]
pawelwilczynski: I moved it up

pawelwilczynski:
I applied your comments and replaced `s6` with `init: true` option. Mind that I added it to docker-compose and I don't know how do you run containers in production but you need to remember about passing `--init` option to `docker run` or other way.

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,41 @@
...
+
+RUN pip3 install --user --no-cache-dir -r /srv/app/requirements.txt
+
+
+{% if cookiecutter.use_alpine_linux == "y" -%}
+FROM python:3.6-alpine AS secondary-image
+{% else -%}
+FROM python:3.6 AS secondary-image
+{%- endif %}
+COPY --from=base-image /root/.local /root/.local
```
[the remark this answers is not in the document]
pawelwilczynski: Ok, moved to `/root`. I hope that existing apps don't rely on `/srv/app/` absolute path anywhere

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,38 @@
...
-RUN tar xvfz /tmp/s6-overlay-amd64.tar.gz -C / \
-    && rm -f /tmp/s6-overlay-amd64.tar.gz {% if cookiecutter.use_celery == "y" %}\
-    && adduser --disabled-password --gecos '' celery{% endif %}
+{% if cookiecutter.use_alpine_linux == "y" -%}
+RUN apk add postgresql-dev gcc python3-dev musl-dev
+{% endif -%}
 
-COPY ./src/ /srv/app/
+ENV PYTHONUNBUFFERED=1
+ENV PATH=/root/.local/bin:$PATH
```
ppolewicz: most Dockerfile instructions, including ENV vars, are scoped per build-stage, and will not be preserved in the next stage

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,38 @@
...
 
-WORKDIR /srv/app
+RUN pip3 install --user --no-cache-dir psycopg2-binary==2.8.3 gunicorn==19.9.0
```
ppolewicz: hardcoded versions. Grep them out from `requirements.txt`

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,38 @@
...
 
-RUN pip3 install --no-cache-dir -r requirements.txt gunicorn\
-    && python3 -m compileall -b .
+WORKDIR /root/app
 
-ENTRYPOINT ["/init"]
+COPY ./src/requirements.txt /root/app/requirements.txt
```
ppolewicz: why the custom path, different directory structure in docker than in repo? couldn't we just use `COPY ./src/requirements.txt src/requirements.txt` and install from there?

{{cookiecutter.repostory_name}}/dc-prod.yml
```diff
@@ -1,26 +1,28 @@
...
     volumes:
       - ./db/data:/var/lib/postgresql/data
 
-
   app:
     build:
       context: ./app
       dockerfile: Dockerfile
     image: {{cookiecutter.django_project_name}}/app
+    init: true
```
ppolewicz: I know init is necessary in `celery-worker` and `app`, but is it needed in others?

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -1,7 +1,12 @@
...
-docker-compose down -v
+PRIMARY_IMAGE_HASH=$(docker-compose build app | tee /dev/tty \
+    | grep "AS secondary-image" -B1 | head -1 | awk '{print $2}')
+docker-compose build  # build the rest of images
+docker-compose down -v --remove-orphans
 docker-compose up -d
 docker-compose exec app python manage.py migrate
 docker-compose exec app python manage.py collectstatic
-docker images --quiet --filter=dangling=true | xargs --no-run-if-empty docker rmi || true
+docker images --quiet --filter=dangling=true \
```
ppolewicz: won't it clean all the layers of primary image except the last one?

{{cookiecutter.repostory_name}}/app/Dockerfile
pawelwilczynski: right

{{cookiecutter.repostory_name}}/app/Dockerfile
pawelwilczynski: Originally it was `COPY ./src/ /srv/app/` . It's not a different directory structure. It's just renaming `src` to `app`. I kept the convention, but I can change it to `src`.

{{cookiecutter.repostory_name}}/dc-prod.yml
pawelwilczynski: From what I know `pg`, `redis` don't spawn any additional subprocesses and aren't started from shell so zombies shouldn't leak but being extra-cautious with `init: true` doesn't cost much, at least I don't see any slow down

{{cookiecutter.repostory_name}}/deploy.sh
pawelwilczynski: No. After building images and running `docker images --quiet --filter=dangling=true` the only dangling image is the final hash of the first stage.

{{cookiecutter.repostory_name}}/app/Dockerfile
pawelwilczynski: So, `gunicorn` hasn't been defined in `requirements.txt` and was never pinned to any version. I also removed `psycopg2` from `requirements.txt` and I don't know if it really needs to exists in there.

{{cookiecutter.repostory_name}}/app/Dockerfile
ppolewicz: lets keep it as simple as we can

{{cookiecutter.repostory_name}}/app/Dockerfile
ppolewicz: we may not need postgres at all if we are setting up a project for a customer that uses mysql or mssql (it sometimes happens)

{{cookiecutter.repostory_name}}/dc-prod.yml
ppolewicz: it costs a little bit of memory and if pg or redis would create zombies, it would be a bug in their apps and they'd have to fix it. Lets not waste memory on an init process that will never do anything.

{{cookiecutter.repostory_name}}/deploy.sh
ppolewicz: That's odd. What if you'd modify requirements, would it then cause more images to be dangling? Will all of them be detected by the first run of `docker images --quiet --filter=dangling=true` or do we need to run it in a loop until it cleans up everything?

{{cookiecutter.repostory_name}}/app/Dockerfile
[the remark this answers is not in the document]
ppolewicz: if they would, that would be a bug (also, that would crash in local development)

{{cookiecutter.repostory_name}}/dc-prod.yml
pawelwilczynski: Removed `init: true` from `redis`, `postgres` and `nginx`. I left it in all celery-related services as they import our django code.

pawelwilczynski:
Further improvements that I comes to my mind:
* Using [tmpfs](https://docs.docker.com/compose/compose-file/#tmpfs) for another postgres instance dedicated for running the tests + setting `-c fsync=off -c full_page_writes=off`. It may speed up creating db, schema, running the tests and tearing them down. I often see ~2 times improvement.
* Using [healthchecks](https://docs.docker.com/compose/compose-file/#healthcheck) so services would start after dependent services is healthy and not just booting up. Also `docker ps` would show if the application is `starting`, `healthy` or `unhealhy` (if stopped responding to the healthcheck)

{{cookiecutter.repostory_name}}/deploy.sh
pawelwilczynski: I think that `--filter=dangling=true` doesn't include all images, just final and stages. I can list all the intermediate images with `docker images -a` but combining `docker images -a --filter=dangling=true` doesn't list intermediate images. On the other hand `docker rmi` removes an image and intermediate images that aren't used any more.

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,38 @@
...
+RUN grep psycopg2 requirements.txt && \
+    apk add postgresql-dev gcc python3-dev musl-dev
+{% endif -%}
 
-ENTRYPOINT ["/init"]
+RUN pip3 install --user --no-cache-dir -r requirements.txt
+
+COPY ./src/ /root/src/
+
+RUN python3 -m compileall -b -f -q .
```
pawelwilczynski: I'm not sure if to compile just django project files or python dependencies installed in `/root/.local` as well?

pawelwilczynski:
I couldn't install the docker-compose from `apt` sources... Needed `1.22.0+` version can be installed only in a way described in https://docs.docker.com/compose/install/#install-compose which is: 
```
$ sudo curl -L "https://github.com/docker/compose/releases/download/1.24.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
$ sudo chmod +x /usr/local/bin/docker-compose
$ sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
```

{{cookiecutter.repostory_name}}/deploy.sh
ppolewicz: ah ok now I get it

{{cookiecutter.repostory_name}}/app/Dockerfile
ppolewicz: my understanding is that whatever is not compiled at that stage, will be compiled in runtime

{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -1,19 +1,38 @@
...
 
-COPY ./bin/s6-overlay-amd64.tar.gz /tmp/s6-overlay-amd64.tar.gz
+FROM $BASE_IMAGE AS base-image
 
-RUN tar xvfz /tmp/s6-overlay-amd64.tar.gz -C / \
-    && rm -f /tmp/s6-overlay-amd64.tar.gz {% if cookiecutter.use_celery == "y" %}\
-    && adduser --disabled-password --gecos '' celery{% endif %}
+WORKDIR /root/src/
+ENV PYTHONUNBUFFERED=1
+ENV PATH=/root/.local/bin:$PATH
```
ppolewicz: do we need those env vars in the first stage?

{{cookiecutter.repostory_name}}/app/Dockerfile
ppolewicz: my understanding is that whatever is not compiled at that stage, will be compiled in runtime

pawelwilczynski: It will, right.
I was just thinking if there's a point in compiling all of the code of dependencies when sometimes you just use 10% of the library functionality. If a few of additional MB added to the image is fine we can compile everything.

{{cookiecutter.repostory_name}}/app/Dockerfile
pawelwilczynski: `PYTHONUNBUFFERED=1` no.
but `PATH=/root/.local/bin:$PATH` helps to avoid warnings printed at dependencies installation saying that you install something (`--user`) out of the `PATH`

PR #19, opened 2019-11-04
deploy.sh made slightly faster (shorter downtime)

pwilczynskiclearcode:


{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN pip3 install --user --no-cache-dir -r requirements.txt
 COPY ./src/ /root/src/
 
 RUN python3 -m compileall -b -f -q /root/
+RUN SECRET_KEY=0 CSP_ENABLED= DATABASE_URL= python3 manage.py collectstatic --no-input --clear
```
pwilczynskiclearcode: By tricking with these env variables we can run `collectstatic` during the container build

{{cookiecutter.repostory_name}}/dc-prod.yml
```diff
@@ -36,8 +36,8 @@ services:
     restart: always
     env_file: ./.env
     volumes:
-      - backend-static:/srv/app/static
-      - ./media:/srv/app/media
+      - backend-static:/root/src/static
+      - ./media:/root/src/media
```
pwilczynskiclearcode: these path should have been changed before, my mistake

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -2,10 +2,18 @@
 PRIMARY_IMAGE_HASH=$(docker-compose build app | tee /dev/tty \
     | sed '/^$/d' | grep "AS secondary-image" -B1 | head -1 | awk '{print $2}')
 docker-compose build  # build the rest of images
-docker-compose down -v --remove-orphans
```
pwilczynskiclearcode: `-v` is for stopping volumes. I don't think we really need to stop them when restarting the services

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -2,10 +2,18 @@
...
+{% if cookiecutter.use_celery == "y" -%}
+    celery-worker celery-beat \
+{%- endif %}
+{% if cookiecutter.use_flower == "y" -%}
+    celery-flower
+{%- endif %}
+
 docker-compose exec app python manage.py migrate
-docker-compose exec app python manage.py collectstatic
+docker-compose exec nginx nginx -s reload
```
pwilczynskiclearcode: I doubt there are many of changes to the `nginx` service declaration in `docker-compose.yml` itself. I believe that most of the changes will be done to the nginx config so reloading it without the process downtime might be much better as it won't terminate connections established between clients and nginx

{{cookiecutter.repostory_name}}/dc-prod.yml
```diff
@@ -11,7 +11,7 @@ services:
     logging: &logging
       driver: journald
       options:
-        tag: '{{.Name}}'
+        tag: {% raw -%}'{{.Name}}'{%- endraw %}
```
pwilczynskiclearcode: without the `{% raw %}` cookiecutter tries to find `.Name` variable in its configuration

pwilczynskiclearcode:
@ppolewicz Could you have a look?

{{cookiecutter.repostory_name}}/deploy.sh
ppolewicz: nope, lets remove the flag if you don't think we need it

{{cookiecutter.repostory_name}}/app/Dockerfile
ppolewicz: sure, but please add it as a comment

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -2,10 +2,26 @@
...
+
+docker-compose stop app \
+{% if cookiecutter.use_celery == "y" -%}
+    celery-worker celery-beat \
+{%- endif %}
+{% if cookiecutter.use_flower == "y" -%}
+    celery-flower
+{%- endif %}
+
+docker-compose start app \
```
ppolewicz: https://stackoverflow.com/a/33716320/95920 says it won't create the container in the first place. Can we use `up`?

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -2,10 +2,26 @@
...
+{% if cookiecutter.use_celery == "y" -%}
+    celery-worker celery-beat \
+{%- endif %}
+{% if cookiecutter.use_flower == "y" -%}
+    celery-flower
+{%- endif %}
```
ppolewicz: is there some way in which we can avoid manually enumerating container names? skip the nginx and do everything else?

{{cookiecutter.repostory_name}}/deploy.sh
ppolewicz: sure, but please add a comment

{{cookiecutter.repostory_name}}/deploy.sh
pwilczynskiclearcode: Yes, actually I don't need to specify services when running `up`. Every single service needs to be up

{{cookiecutter.repostory_name}}/deploy.sh
pwilczynskiclearcode: So I resigned from jinja variables in favour of listing services using `docker-compose ps`

pwilczynskiclearcode:
@ppolewicz Could you look at the changes once again?

PR #23, opened 2020-01-21
Add dbshell.sh

ppolewicz:


pawelpolewicz:
I'm merging it how it is and we'll update later if needed

PR #29, opened 2020-05-15
Bump versions, use py38

mlech-reef:


{{cookiecutter.repostory_name}}/app/Dockerfile
```diff
@@ -36,6 +36,8 @@ ENV PATH=/root/.local/bin:$PATH
 
 COPY --from=base-image /root/ /root/
 
+RUN grep psycopg2 requirements.txt && apk add --no-cache libpq || true
```
ppolewicz, line 39: this should only happen on alpine, right?

mlech-reef: Oh, yeah. Didn't surface during testing as `|| true` did the trick, but a more elegant solution is to add `if`.

PR #47, opened 2021-08-13
Fixes #43

### LINKED ISSUE #43: Fix docker-compose health check of the app and nginx services
Use e.g. `HEAD /admin/login/` which should be always available in opposite to `GET /`

agoncharov-reef:
* `|| exit 1` is needed because docker-compose says that exit code may be 0 (ok), 1 (error) or 2 (reserved).
* added celery healthcheck
* `--spider` makes HEAD request instead of GET


mszumocki-reef:
I've noticed in EST that GET / health check creates tons of index.html.<number> files in root dir of Docker container. 
This one will not have the same problem I hope?

PR #48, opened 2021-08-16
One small fix (redis host) and a minor feature (adding the django app to virtualenv)

mpnowacki-reef:


{{cookiecutter.repostory_name}}/setup-dev.sh
```diff
@@ -16,6 +16,10 @@ pip install --upgrade -r "${PROJECT_DIR}/app/src/requirements.txt"
 # Create .env from the template if doesn't exist
 [[ -f "${ENV_DIR}/.env" ]] || cp "${ENV_DIR}/.env.template" "${ENV_DIR}/.env"
 
+# "install" this package in the venv
+# works for python versions
+find "$VIRTUAL_ENV/lib/" -type d -maxdepth 1 -mindepth 1 -exec bash -c 'echo "$PWD/app/src" > "$1/site-packages/this.pth"' bash {} \;
```
mszumocki-reef, line 21: Why is this needed?

mpnowacki-reef: without this, my IDE doesn't properly find packages, like `proj.app.models` when using absolute imports. With this, it does not warn about missing package (this is purely cosmetic, as the code runs) and resolves symbols properly.

mszumocki-reef: That sounds like incorrect project configuration in the IDE (wrong source root perhaps)? (works fine in PyCharm for me without it)

mpnowacki-reef: When starting a new project, do you simply open the repo root and choose an interpreter? Or do you do any additional steps to have it find all the app packages?

mszumocki-reef: I set the correct venv and the go to settings -> project structure and mark app/src as source folder.

mpnowacki-reef: I see. Okay, I guess that's the better way, requires a one-time action on every project, but has less magic

PR #52, opened 2021-10-11
Add redirect 80->443 when use_https=y; clean up nginx https-related template

agoncharov-reef:
This adds http -> https redirect; removed commented sections (we may always copy them from cookiecutter template when needed)

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -4,7 +4,6 @@
     "django_project_name": "project",
     "django_default_app_name": "core",
     "postgres_user": "postgres",
-    "use_https": "y",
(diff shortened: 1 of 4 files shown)


ppolewicz:
I suggest that we remove the https switch and leave it always on. If someone needs it to be disabled, that's going to be a special request and we'll send a devops engineer to figure it out.

agoncharov-reef:
Having https switch always ON won't allow deploy a project without a domain. We didn't have a domain for EST for a while.
Do you really want to exclude an option of "plain IP address" for early deployment?

PR #55, opened 2021-10-29
Run manage.py in standalone container

agoncharov-reef:
Currently `run-manage-py.sh` command uses `docker-compose exec` to run `manage.py` in the same container where the app is running.
Proposed change runs command in a separate container, so the container inherits the same environment as running app, but does not interfere with any processes of the app and is not dismissed even during application redeployment - so long-running management command and shell sessions are not interrupted.


ppolewicz:
The new behavior should be optional and not the default

agoncharov-reef:
* `bin/run-manage-py.sh command`
* `bin/run-manage-py.sh detached command`
@ppolewicz ?

ppolewicz:
bin/run-manage-py.sh command
bin/run-manage-py-detached.sh command

PR #60, opened 2021-11-30
added monitoring

mpnowacki-reef:


{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -118,14 +118,57 @@ services:
...
+      - '--path.rootfs=/rootfs'
+      - '--path.sysfs=/host/sys'
+      - '--collector.filesystem.mount-points-exclude=^/(sys|proc|dev|host|etc)($$|/)'
+    logging:
+      <<: *logging
+
+  cadvisor:
+    image: gcr.io/cadvisor/cadvisor:v0.40.0
+    container_name: cadvisor
+    privileged: true
```
agoncharov-reef, line 158: Isn't this sorta dangerous? I quickly googled and here they don't use `privileged` - https://prometheus.io/docs/guides/cadvisor/

{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
```diff
@@ -134,3 +134,63 @@ server {
...
+    server_name monitoring;
+    listen      10443           ssl http2;
+
+    ssl_session_cache         shared:SSL:20m;
+    ssl_session_timeout       10m;
+
+    ssl_protocols             TLSv1 TLSv1.1 TLSv1.2;
+    ssl_prefer_server_ciphers on;
+    ssl_ciphers               "ECDH+AESGCM:ECDH+AES256:ECDH+AES128:!ADH:!AECDH:!MD5;";
+
```
agoncharov-reef, line 149: I think you can copy this from other parts of nginx config, or create some common config file and import it in both places

{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
```diff
@@ -134,3 +134,63 @@ server {
...
+    ssl_trusted_certificate   /etc/monitoring_certs/ca.crt;
+
+    ssl_client_certificate /etc/monitoring_certs/ca.crt;
+    ssl_verify_client on;
+
+    resolver 8.8.8.8 8.8.4.4 valid=300s;
+    resolver_timeout 5s;
+
+
+    gzip off;
```
agoncharov-reef, line 164: I think you can copy this from other parts of nginx config, or create some common config file and import it in both places

{{cookiecutter.repostory_name}}/nginx/monitoring_certs/README
```diff
@@ -0,0 +1,2 @@
+Obtain ssl cert and key for this deployment and place them here, named `cert.crt` and `cert.key`. Also get the ca cert
+used to verify the client fetching monitoring data and place it here, name `ca.crt`.
```
agoncharov-reef, line 2: Sorry, with this instruction I won't be able to deploy myself. I don't get where I get the certificates from :(

agoncharov-reef (changes requested):

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
mpnowacki-reef: in some (several, actually) cases it's needed: https://github.com/google/cadvisor/blob/master/docs/running.md

but let's turn it off and see if it breaks somewhere :)

pawelpolewicz (approved):

PR #61, opened 2022-01-08
Install AWS CLI and jq tool in prepare-os.sh

awierzbicki-reef:
They're needed to use custom Python package index hosted on AWS.

### DIFF
--- a/{{cookiecutter.repostory_name}}/bin/prepare-os.sh
+++ b/{{cookiecutter.repostory_name}}/bin/prepare-os.sh
@@ -13,6 +13,12 @@ if [ -x "${DOCKER_BIN}" ] && [ -x "${DOCKER_COMPOSE_BIN}" ] && [ -x "${SENTRY_CL
     exit 1;
 fi
 
+PLATFORM="$(uname -i)"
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repostory_name}}/bin/prepare-os.sh
```diff
@@ -35,3 +48,14 @@ fi
 if [ ! -x "${DOCKER_COMPOSE_BIN}" ]; then
     apt-get -y install docker-ce docker-compose
 fi
+
+if [ ! -x "${AWS_CLI}" ]; then
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
ppolewicz, line 53: please add support for ARM and consistency verification. We will soon need to experiment with various kinds of hardware to measure efficiency of a 3rd party computation on various (aws) platforms, not being able to run it on ARM would be a disadvantage.

ppolewicz (approved):

awierzbicki-reef:
@ppolewicz Seems like I lack permissions to push docker image. I made this PR by forking the repo since I could do it another way.

ppolewicz:
Then CI should not try to push in that situation

PR #63, opened 2022-01-12
AWS cd

mpnowacki-reef:


{{cookiecutter.repostory_name}}/.github/workflows/cd.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  push:
+    branches:
+      - 'deploy-*'
+
+jobs:
+  deploy:
+    env:
+      AWS_ACCESS_KEY_ID: {% raw %} ${{ secrets.DEPLOYMENT_AWS_ACCESS_KEY_ID }} {% endraw %}
+      AWS_SECRET_ACCESS_KEY: {% raw %} ${{ secrets.DEPLOYMENT_AWS_SECRET_ACCESS_KEY }} {% endraw %}
```
agoncharov-reef, line 12: `deploy-to-aws.sh` contains other env vars as well:
```
APP_OWNER="12345"
APP_REGION="us-east-1"
APP_NAME="trololo${APP_PREFIX}"
```
Where do they come from?

mpnowacki-reef: They are defined in deploy-to-aws.sh. The repositories are not deployment-agnostic, terraform stuff contains AWS account names and domains.

agoncharov-reef: So you say that for each project we gonna modify variables in `deploy-to-aws.sh` and `tf-main/...`? I thought that we wanted to have only typical configuration in cookiecutter, so that we only have to modify `.env` files...

mpnowacki-reef: That’s the plan. It’s been scheduled to do by one devops engineer who is currently caught up in other stuff

mpnowacki-reef: It doesn’t affect cd.yml, though

ppolewicz (approved):

PR #73, opened 2022-06-08
Constance metrics

msalaban-reef:
This PR allows for exporting `django-constance` config variables as Prometheus metrics.

1. It supports `int`, `float`, `bool` and `decimal.Decimal` types by default.
2. It skips variables listed in `CONSTANCE_PROMETHEUS_BLACKLIST` setting.
3. It processes variables listed in `CONSTANCE_PROMETHEUS_WHITELIST` setting by adding their string representation to the metric's name. (May cause problems in multiprocess mode.)

This is a review request, esp. to @ppolewicz


{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+    if gets_monitored(key, new_value):
+        if key not in METRICS:
+            METRICS[key] = Metric(key)
+        METRICS[key].store(new_value)
+
+
+def export_config():
+    for key in dir(config):
+        value = getattr(config, key)
+        if gets_monitored(key, value):
```
kdowolski-reef, line 118: Would be nice to write function like **store_monitored{,_metric}**

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+class Metric:
+    prefix = "constance_config"
+    config_key = None
+
+    def __init__(self, config_key=None):
+        self.config_key = config_key or self.config_key
+        self.name = self._get_name()
+        self.description = constance_settings.CONFIG[self.config_key][1]
+        value = getattr(config, self.config_key)
+        if isinstance(value, (bool, int, float, Decimal)):
```
ppolewicz, line 59: this is duplicated in line 43, please move this logic to a function such as `is_of_whitelisted_type(value)`, so that if we ever want to add something, there is no possibility of someone missing the fact that the type needs to be added in two places in the code.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+        isinstance(value, (int, float, bool, Decimal)) or is_whitelisted(key)
+    )
+
+
+METRIC_NAME_FILTER = re.compile(r"\W", re.ASCII)
+
+
+class Metric:
+    prefix = "constance_config"
+    config_key = None
```
ppolewicz, line 52: this class-level variable is shadowed by a field in constructor. This can cause trouble in many ways, while the default is declared in the __init__ signature. Please remove it from the class scope and line 55 can just be `self.config_key = config_key` (`None` by default if the caller doesn't pass the name in)

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+    return not is_blacklisted(key) and (
+        isinstance(value, (int, float, bool, Decimal)) or is_whitelisted(key)
+    )
+
+
+METRIC_NAME_FILTER = re.compile(r"\W", re.ASCII)
+
+
+class Metric:
+    prefix = "constance_config"
```
ppolewicz, line 51:
```suggestion
    PREFIX = "constance_config"
```
as per [PEP8](https://peps.python.org/pep-0008/#constants)

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+            self.store = self._metric.set
+        else:
+            # a str or custom type; cast everything to str
+            value = str(value)
+            self.name = self._get_name(value)
+            self._metric = Gauge(self.name, self.description)
+            self.store = self._store_str
+
+    def _get_name(self, value=None):
+        n = f"{self.prefix:s}_{self.config_key:s}"
```
ppolewicz, line 70: While typing a single letter variable saves a name completion keystroke, please observe how confusing they are for people who try to debug something that doesn't work right at `03:25 AM`. It's a `name_candidate`, right?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+        if value is not None:
+            v = METRIC_NAME_FILTER.sub('_', value)
+            n = f"{n}_{v}"
+        return n
+
+    def _unregister(self):
```
ppolewicz, line 76: @mpnowacki-reef I think you might know what to do here. I mean for me it's pretty clear that unregistering should be implemented in `metrics.py` and not in `constance_prometheus.py`, registry seems to be a singleton so ideally we'd see some CollectorRegistryProvider which would just return the exiting object, rather than the class magically returning the same object as created somewhere else or the class somehow working consistently across threads despite having multiple copies...

But other than that I don't know how to deal with the multiprocess unregistering.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+from django.conf import settings
+from django.dispatch import receiver
+from prometheus_client import REGISTRY, CollectorRegistry, Gauge
+
+from . import metrics
+
+__all__ = ("Metric", "on_constance_config_updated", "export_config")
+
+
+def is_blacklisted(key):
```
ppolewicz, line 33: function signature type hints are missing in the entire module

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+            # NOTE: This will fail in multiprocess mode
+            registry.unregister(self._metric)
+        except KeyError:
+            # FIXME: Once unregistering fails, the metric remains in the registry.
+            # As the value of string variables is converted to the metric name,
+            # returning to previous setting may cause duplicates to appear.
+            pass
+
+    def _store_str(self, value):
+        value = str(value)
```
ppolewicz, line 92: make the caller pass the correct type to the function

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+from constance import config, settings as constance_settings
+from constance.signals import config_updated
+from decimal import Decimal
+from django.conf import settings
+from django.dispatch import receiver
+from prometheus_client import REGISTRY, CollectorRegistry, Gauge
+
+from . import metrics
+
+__all__ = ("Metric", "on_constance_config_updated", "export_config")
```
ppolewicz, line 30: Why do this?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,124 @@
...
+        METRICS[key].store(new_value)
+
+
+def export_config():
+    for key in dir(config):
+        value = getattr(config, key)
+        if gets_monitored(key, value):
+            if key not in METRICS:
+                METRICS[key] = Metric(key)
+            METRICS[key].store(value)
```
ppolewicz, line 121: direct clone of lines 109-112, please refactor

cookiecutter.json
```diff
@@ -8,6 +8,9 @@
     "postgres_user": "postgres",
     "use_celery": "y",
     "use_flower": "{% if cookiecutter.use_celery == 'y' %}y{% else %}n{% endif %}",
+    "use_constance": "y",
+    "constance_backend": "database",
+    "constance_redis_url": "{% if cookiecutter.constance_backend == 'redis' %}redis://localhost:6379/0{% endif %}",
```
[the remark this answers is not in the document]
msalaban-reef: I did.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
msalaban-reef: The caller would have to do another type check then, which would make the code unnecessarily complex.
ATM the only additional types allowed by constance are those from `datetime` and `str` works on them pretty well.

msalaban-reef:
Another idea here is to keep the following types permitted by constance as:
* `datetime`: `.totimestamp()`, `.fromtimestamp()`
* `date`: `.toordinal()`, `fromordinal()`
* `timedelta`: `td / timedelta(microseconds=1)`
* `time`: in microseconds, but the conversion methods aren't provided explicitly

The advantage is that we could keep them as `float`s, at the cost of additional conversion on both ends. Problems arise, however, when timezones are used and need to be conserved (otherwise we could normalize to UTC).

@ppolewicz what do you think about it?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
mpnowacki-reef: registry seems to not be a singleton, according to prometheus-client howto: https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn


I'm a bit lost, it is my understanding that `registry.unregister` is for unregistering Collectors, not individual metrics, what ma I missing?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
```diff
@@ -0,0 +1,133 @@
...
+        self.description = constance_settings.CONFIG[self.config_key][1]
+        value = getattr(config, self.config_key)
+        if is_of_automatically_monitored_type(value):
+            self._metric = Gauge(self.name, self.description)
+            self.store = self._metric.set
+        else:
+            # a str or custom type; cast everything to str
+            value = str(value)
+            self.name = self._get_name(value)
+            self._metric = Gauge(self.name, self.description)
```
mpnowacki-reef, line 68: I'm worried about the `multiprocess_mode` being let default here, which is `all`. I think what we need is `livemax`. The current behaviour will be:

'all': Default. Return a timeseries per process (alive or dead), labelled by the process's pid (the label is added internally).

If constance yields different values in different processes, it's either a misconfiguration or a bug, not sure we should be detecting it here

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
mpnowacki-reef: My question is: does it actually work? When collecting metrics in multiprocessing mode, their values are read from .db files. I don't se how `unregister` would remove entries from those files but I can be wrong. 

Frankly speaking, from what I see, removing metrics does not really work when in multiprocessing mode. But again, I could be wrong. This whole thing is a little bit spaghetti in prometheus-client. Removing labels, however, should work, so a single metric with labels for constance param name and value could be a quasi-supported solution.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
msalaban-reef: It doesn't help there, either. Adding the mode, I still get a duplicate error when changing the variable back to previously used value:

```
Environment:


Request Method: POST
Request URL: http://localhost:8000/admin/constance/config/

Django Version: 3.2.12
Python Version: 3.10.4
Installed Applications:
['django_prometheus',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django_extensions',
 'django_probes',
 'constance',
 'project.core',
 'corsheaders']
Installed Middleware:
['corsheaders.middleware.CorsMiddleware',
 'django_prometheus.middleware.PrometheusBeforeMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django_prometheus.middleware.PrometheusAfterMiddleware']



Traceback (most recent call last):
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/contrib/admin/sites.py", line 232, in inner
    return view(request, *args, **kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/constance/admin.py", line 256, in changelist_view
    form.save()
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/constance/admin.py", line 157, in save
    setattr(config, name, new)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/utils/functional.py", line 277, in __setattr__
    setattr(self._wrapped, name, value)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/constance/base.py", line 29, in __setattr__
    self._backend.set(key, value)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/constance/backends/redisd.py", line 50, in set
    signals.config_updated.send(
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 180, in send
    return [
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 181, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/home/emes/devel/reef/cct/project/app/src/project/core/constance_prometheus.py", line 134, in on_constance_config_updated
    store_monitored_metric(key, new_value)
  File "/home/emes/devel/reef/cct/project/app/src/project/core/constance_prometheus.py", line 117, in store_monitored_metric
    METRICS[key].store(new_value)
  File "/home/emes/devel/reef/cct/project/app/src/project/core/constance_prometheus.py", line 106, in _store_str
    self._metric = Gauge(self.name, self.description)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/prometheus_client/metrics.py", line 365, in __init__
    super().__init__(
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/prometheus_client/metrics.py", line 143, in __init__
    registry.register(self)
  File "/home/emes/venvs/reef/lib/python3.10/site-packages/prometheus_client/registry.py", line 43, in register
    raise ValueError(

Exception Type: ValueError at /admin/constance/config/
Exception Value: Duplicated timeseries in CollectorRegistry: {'constance_config_SOME_STRING_hello_string'}
```

I wonder if we shouldn't drop the `str` (and other non-standard type) monitoring until we figure out a way to handle it better.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
msalaban-reef: I think it's kind of fixed right now. I say _kind_ because 15cd1fc is nothing but a hack. It seems to work in my local test project but will likely bite our collective ass in the future. What do you think?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/constance_prometheus.py
msalaban-reef: Nevermind the above comment. I'll address the `"livemax"` mode later on.
BTW, have you noticed that `"livemax"` is available only in unreleased code? Latest official release (0.14.1) doesn't have it.

msalaban-reef:
Any progress with reviewing this, guys?

PR #75, opened 2022-06-11
Add sending slack notification on deploy

kdowolski-reef:


{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -27,6 +27,11 @@ docker-compose run --rm app sh -c "python manage.py wait_for_database --timeout
 # start everything
 docker-compose up -d
 
+{% if cookiecutter.use_slack_notification %}
+# Send slack notification about deploy
+docker-compose run --rm app sh -c "python bin/notify.py"
```
kdowolski-reef, line 32: Not sure how working with that template looks like in real but **deploy-to-aws** (which is called by github action) doesn't call **deploy** (or I don't see something). If so, the problem where to call it stands.

{{cookiecutter.repostory_name}}/.github/workflows/cd.yml
```diff
@@ -22,4 +22,7 @@ jobs:
           ./deploy-to-aws.sh $ENVIRONMENT
           export TAG=deployed-${ENVIRONMENT}-`date -u +"%Y-%m-%dT%H.%M.%S"`
           git tag $TAG
-          git push origin $TAG
\ No newline at end of file
+          git push origin $TAG
+        env:
+          commit_msg : {% raw %} ${{ github.event.head_commit.message }} {% endraw %}
```
kdowolski-reef, line 27: Could change that for list of all commit messages since last deploy but need to do some testing for that

cookiecutter.json
```diff
@@ -38,5 +38,7 @@
     "aws_base_domain_name": "fake-domain.com",
     "aws_domain_name": "api.{{ cookiecutter.aws_base_domain_name }}",
     "aws_staging_domain_name": "staging.{{ cookiecutter.aws_domain_name }}",
-    "aws_ec2_ssh_key": ""
+    "aws_ec2_ssh_key": "",
+    "use_slack_notifications": "y",
+    "slack_channel": "{% if cookiecutter.use_slack_notification == 'y' %}#develop{% else %}{% endif %}"
```
ppolewicz, line 43: have you tested it? Looks like chicken-and-egg problem

{{cookiecutter.repostory_name}}/.github/workflows/cd.yml
ppolewicz: we need to see all commits that have been deployed

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
+import argparse
+import os
+from datetime import datetime
+
+import requests
+
+msg_template = "*{timestamp}*\n{author}\n{changelog}"
```
ppolewicz, line 7: https://peps.python.org/pep-0008/#constants

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class SlackNotifier:
+    url = "https://slack.com/api/chat.postMessage"
```
ppolewicz, line 11: https://peps.python.org/pep-0008/#constants

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, token, channel):
+        if not channel or not token:
+            raise Exception("Missing token or channel")
```
ppolewicz, line 15: when would that happen?

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+
+import requests
+
+msg_template = "*{timestamp}*\n{author}\n{changelog}"
+
+
+class SlackNotifier:
+    url = "https://slack.com/api/chat.postMessage"
+
+    def __init__(self, token, channel):
```
ppolewicz, line 13: missing type hints

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+        if not channel or not token:
+            raise Exception("Missing token or channel")
+
+        self.token = token
+        self.channel = channel
+
+    def notify(self, text):
+        print(f"Sending message to channel {self.channel}")
```
ppolewicz, line 21: use logger instead of printing to stdout

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+        r = requests.post(
```
ppolewicz, line 22: as this class can be used more than 2 times per minute (not in this PR but in the future), make requests.Session in __init__ and use it here to prevent breaking a https connection after every request

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+            self.url,
+            data={
+                "text": text,
+                "channel": self.channel,
+                "token": self.token,
+            },
+        )
+        r.raise_for_status()
```
ppolewicz, line 30: single letter variables can save you a variable name completion keystroke but make things harder for engineers who will try to fix something that doesn't work right in the middle of the night. It's `response`, lets call it that way.

{{cookiecutter.repostory_name}}/bin/notify.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def get_deployment_author():
+    return os.environ.get("sender") or "Deployed manually"
+
+
+def get_deployment_changelog():
+    # for now it's just message of last commit when deployed with github action
```
ppolewicz, line 38: it doesn't work if it's not a github action and it doesn't work at all if the commit is "merged branch xxx to yyy" because we still don't know what changes have been deployed

{{cookiecutter.repostory_name}}/deploy.sh
ppolewicz: What solution do you suggest?

kdowolski-reef: I see two options.
1) Run it [here](https://github.com/reef-technologies/cookiecutter-rt-django/blob/master/%7B%7Bcookiecutter.repostory_name%7D%7D/devops/scripts/deploy-backend.sh#L10). Technically we have container built there so could run script. Currently we do not COPY **bin** directory to container but I guess it's not intended.
2) Technically we could add it [here](https://github.com/reef-technologies/cookiecutter-rt-django/blob/master/%7B%7Bcookiecutter.repostory_name%7D%7D/app/envs/prod/entrypoint.sh#L7) but in case of restarting container we will get notifications.

{{cookiecutter.repostory_name}}/deploy.sh
```diff
@@ -29,7 +29,8 @@ docker-compose up -d
 
 {% if cookiecutter.use_slack_notification %}
 # Send slack notification about deploy
-docker-compose run --rm app sh -c "python bin/notify.py"
+output=`git log --format=format:%H,%s`
+docker-compose run --rm app sh -c "python /root/src/bin/notify.py --parse -m \"$output\""
```
kdowolski-reef, line 33: Unfortunately it allows for code injection with backticks, searching how to escape it properly.

PR #81, opened 2022-08-12
Only import debug_toolbar when needed

vbaltrusaitis-reef:
`debug_toolbar` version 3.2.1 causes a memory leak in one of our projects. This leak happens even if the toolbar is never used, never added to `urlpatterns` and `DEBUG` is set to `False`. The import is enough.

Newer toolbar versions don't cause this particular problem, but I'm still making the import conditional just in case.

See:
https://github.com/vbaltrusaitis-reef/memleak-demo
https://github.com/jazzband/django-debug-toolbar/issues/906

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
@@ -11,7 +11,6 @@ urlpatterns = [
 ]
 
 if settings.DEBUG_TOOLBAR:
-    import debug_toolbar
     urlpatterns += [
-        path('__debug__/', include(debug_toolbar.urls)),
+        path('__debug__/', include('debug_toolbar.urls')),
     ]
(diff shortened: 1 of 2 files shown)


mpnowacki-reef (approved):

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
```diff
@@ -12,6 +11,7 @@
 ]
 
 if settings.DEBUG_TOOLBAR:
+    import debug_toolbar
     urlpatterns += [
         path('__debug__/', include(debug_toolbar.urls)),
```
matthiask, line 16:
```suggestion
        path('__debug__/', include("debug_toolbar.urls")),
```

You should be able to change the `include` to use a string instead, then you don't need to import the debug toolbar at all.

ppolewicz: If the string passed to `include()` represents a path that cannot be imported (because it's spelled incorrectly, for example), will the exception trigger eagerly during application start or lazily, during the first call that matches the url pattern?

matthiask: I believe Django imports all URLconf entries when starting up.

vbaltrusaitis-reef: @agoncharov-reef  Sure.

PR #83, opened 2022-08-18
Set up dependabot

agoncharov-reef:
This sets up dependabot both for cookiecutter project itself, and for all materialized projects as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/{{cookiecutter.repostory_name}}/app/src"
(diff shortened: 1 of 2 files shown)


mpnowacki-reef:
What happens if a `requirements.txt` file contains a package from a private pypi? It says here https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem that `package-ecosystem=pip` supports private registries, but how to make it work? Provide some sort of credentials somehow somewhere?

mpnowacki-reef:
ah I see we can use `ignore` for that, never mind

ppolewicz (approved):

PR #84, opened 2022-08-19
node-exporter image

mpnowacki-reef:


docker/node-exporter/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+set -x
+
+INSTANCE_TYPE=`wget http://169.254.169.254/latest/meta-data/instance-type -O- --timeout=5`
+
```
ppolewicz, line 7:
```suggestion
set -x

INSTANCE_TYPE_EC2=`wget http://169.254.169.254/latest/meta-data/instance-type -O- --timeout=5`
INSTANCE_TYPE="$(INSTANCE_TYPE_EC2:-UNKNOWN)"
```
lets not just leave the value from the last run (where the instance type might have changed since the last time that container was started)

mpnowacki-reef: I wasn't leaving the value from last run. thanks to `set -e`, if `wget` fails, then the container does not get up. Without it, in case of `wget` failing, alerts depending on that metric will just be ignored silently. The solution you're proposing requires us to add an alert to fire if `node_aws_ec2_instance_type{instance_type="UNKNOWN"} == 1`, is that it? It's a bit fragile, but if fetching instance type fails, at least we'll have all the other node metrics. Tough decision.

ppolewicz: I'd rather have a node up and an alert to deal with than a service disruption because a little monitoring script didn't run as well as it should have.

Also please note that my design replaces an empty 200 response with an UNKNOWN

PR #85, opened 2022-08-22
Add business metrics

vbaltrusaitis-reef:


mpnowacki-reef:
Two buts:
- I wouldn't add a cookiecutter question. too many of those can severely reduce readability. If monitoring is enabled, just add the business metrics as well. If we don't need it, we can just not scrape it
- NGINX config that listens on 8000 needs a block analogous to
```
{% if cookiecutter.monitoring == 'y' %}
    location /metrics {
        return 404;
    }
    {% endif %}
```
otherwise anyone can access the metrics freely

mpnowacki-reef (changes requested):
as stated in my previous comment

vbaltrusaitis-reef:
Addressed the comments (thanks for the catch with nginx! that was a bad one) in the last commits. Also, bumped `django-business-metrics` to version `0.1.0`.

mpnowacki-reef (approved):

PR #92, opened 2022-09-04
Prometheus metric files fix

mlesniewski-reef:
This PR fixes clean ups of Prometheus metrics if they're used in celery workers. Workers, which exit will have their private live metrics removed.

Additionally fixes a few other issues are covered by this MR too:
* Fix for the health check command for the celery-worker container
* Fix for celery-worker container hanging running after celery crashes within the container
* Fix for races between containers related to cleaning up and accessing the metrics volume

See commit messages of individual commits for more details.


mpnowacki-reef:
Celery-worker should have the volume and env var for monitoring set

mlesniewski-reef:
> Celery-worker should have the volume and env var for monitoring set

This is now covered too.

PR #95, opened 2022-10-04
Additional options for sending emails

kkalinowski-reef:
- CC added as a parameter;
- body used to be unused, now either template or provided body is used;
- check that only one of them is available.

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
@@ -26,21 +26,18 @@ def create_attachment(path: str, mime_type: MIMEType = MIMEImage) -> MIMEType:
 
 
 def send_mail(
+    template_name: str,
     subject: str,
     to: list[str],
     from_: str = f'<{settings.DEFAULT_FROM_EMAIL}>',
-    cc: Optional[list[str]] = None,
-    template_name: Optional[str] = None,
-    body: Optional[str] = None,
     context: Optional[dict] = None,
     attachments: Optional[list[str]] = None,
+    cc: Optional[list[str]] = None,
 ):
-    # Ensure that either body or template name is provided, not both.
-    assert (body is not None and template_name is None) or (body is None and template_name is not None)
     context = context or {}
     attachments = attachments or []
 
-    html = body or loader.render_to_string(template_name, context)
+    html = loader.render_to_string(template_name, context)
 
     message = EmailMessage(
         subject=subject,
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
@@ -26,24 +26,28 @@ def create_attachment(path: str, mime_type: MIMEType = MIMEImage) -> MIMEType:
 
 
 def send_mail(
-    template_name: str,
     subject: str,
-    body: str,
     to: list[str],
     from_: str = f'<{settings.DEFAULT_FROM_EMAIL}>',
+    cc: Optional[list[str]] = None,
+    template_name: Optional[str] = None,
+    body: Optional[str] = None,
     context: Optional[dict] = None,
     attachments: Optional[list[str]] = None,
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
```diff
@@ -26,24 +26,28 @@ def create_attachment(path: str, mime_type: MIMEType = MIMEImage) -> MIMEType:
...
     subject: str,
-    body: str,
     to: list[str],
     from_: str = f'<{settings.DEFAULT_FROM_EMAIL}>',
+    cc: Optional[list[str]] = None,
+    template_name: Optional[str] = None,
+    body: Optional[str] = None,
     context: Optional[dict] = None,
     attachments: Optional[list[str]] = None,
 ):
```
vbaltrusaitis-reef, line 37: This is a breaking change if the consumer passes args by position. While one should review template changes when merging, it'd be good to avoid breaking changes if they're not truly necessary.
Though perhaps this is acceptable if we check which of our project use this utility and fix it immediately?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
```diff
@@ -26,24 +26,28 @@ def create_attachment(path: str, mime_type: MIMEType = MIMEImage) -> MIMEType:
...
+    # Ensure that either body or template name is provided, not both.
+    assert (body is not None and template_name is None) or (body is None and template_name is not None)
```
vbaltrusaitis-reef, line 39: It's a shame Python doesn't have an exclusive or. :D

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
kkalinowski-reef: I have no idea about other projects using this, and this is a relatively fresh change (1 week). @agoncharov-reef – as the original creator, maybe you know of a place that currently uses this code?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
kkalinowski-reef: I could go with something like
```python
has_body = body is not None
has_template = template_name is not None
assert has_body != has_template
```
Do you find it more suiting?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
agoncharov-reef: Yeah, my bad - forgot to remove the `body`. I see you left it. Do we need it at all? I think it's a good thing to *force* people to use templates instead of putting email body right in python code.

kkalinowski-reef: That's a fair point. I just wanted to be lazy in RAT, as there is no template and the whole HTML is rendered inside the code (and is as plain as it gets). But I can just remove the body, simplify the rest and roll with it. The only real thing I needed from it, was the `cc` parameter.

kkalinowski-reef: I'll remove the body altogether then.

kkalinowski-reef: Done.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
kkalinowski-reef: Removed, as it wasn't relevant any more.

PR #96, opened 2022-10-06
Don't expose :8000 on nginx

agoncharov-reef:
Screenshot_2022-10-06_12-04-15


vbaltrusaitis-reef:
This will probably be an invalid `docker-compose.yml` if `cookiecutter.monitoring != 'y'`. `ports` must be an array.

PR #105, opened 2023-05-01
Fix node-exporter network stats, remove links

mlech-reef:


mlech-reef:
Having copy-pasted nginx config and docker-compose (a special for AWS) is super annoying :P

ppolewicz (approved):

PR #108, opened 2023-05-20
SemBr using readable

mjurbanski-reef:


{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -60,6 +61,19 @@ def security_check(session):
...
 
+@nox.session(python=PYTHON_DEFAULT_VERSION)
+def readable(session):
+    session.run(
+        'docker',
+        'run',
+        '-v', f'{ROOT.absolute()}:/data',
+        '-w', '/data',
+        '-u', f'{os.geteuid()}:{os.getegid()}',
+        'ghcr.io/bobheadxi/readable:v0.4.0',
```
ppolewicz, line 72: can you put a hash here so that when bobheadxi is hacked and releases a v0.4.0.1 or something with a rootkit, we won't install it here?

ppolewicz (approved):

PR #114, opened 2023-06-10
check if cookiecutter template is valid under CI

mjurbanski-reef:


{{cookiecutter.repostory_name}}/README.md
```diff
@@ -113,14 +113,11 @@ If one wants to deploy other branch, force may be used to push desired branch to
 ```sh
 git push --force production local-branch-to-deploy:master
 ```
-
```
mjurbanski-reef, line 116: changed to make `readable` tool happy with the cruft output

{{cookiecutter.repostory_name}}/nginx/monitoring_certs/README.md
```diff
@@ -1,4 +1,2 @@
-Go to [promehtues-grafana-monitoring](https://github.com/reef-technologies/prometheus-grafana-monitoring) and generate
```
mjurbanski-reef, line 1: changed to make `readable` tool happy with the cruft output

noxfile.py
```diff
@@ -0,0 +1,110 @@
...
+
+@contextlib.contextmanager
+def crufted_project(session):
+    session.run("pip", "install", "-e", ".")
+    global CRUFTED_PROJECT_TMPDIR
+    if not CRUFTED_PROJECT_TMPDIR:
+        CRUFTED_PROJECT_TMPDIR = tempfile.TemporaryDirectory(prefix="rt-crufted_")
+        session.log("Creating project in %s", CRUFTED_PROJECT_TMPDIR.name)
+        session.run("cruft", "create", ".", "--output-dir", CRUFTED_PROJECT_TMPDIR.name, "--no-input")
+        session.notify('cleanup_crufted_project')
```
agoncharov-reef, line 69: I'm not a huge fan of globals... wdyt about `@cache`?


```python
from functools import cache

@cache
def cruft_project() -> Path:
    ...

@contextlib.contextmanager
def crufted_project(session):
    session.run("pip", "install", "-e", ".")
    crufted_project_path = cruft_project()
    project_path = Path(crufted_project_path.name) / "project"
    ...

@nox.session(python=PYTHON_DEFAULT_VERSION)
def cleanup_crufted_project(session):
    if not cruft_project.cache_info().currsize:
        return

    crufted_project_path = cruft_project()
    # workaround for docker-compose creating root-owned files
    rm_root_owned(session, crufted_project_path.name)
    crufted_project_path.cleanup()
    cruft_project.cache_clear()
```

noxfile.py
```diff
@@ -0,0 +1,110 @@
...
+    project_path = Path(CRUFTED_PROJECT_TMPDIR.name) / "project"
+    session.chdir(project_path)
+    yield project_path
+    session.chdir(ROOT)
+
+
+def rm_root_owned(session, dirpath):
+    assert not ROOT.is_relative_to(dirpath)  # sanity check before we nuke dirpath
```
agoncharov-reef, line 77: I see an experienced developer :D

noxfile.py
```diff
@@ -0,0 +1,110 @@
...
+    session.run("docker", "run", "--rm", "-v", f"{dirpath}:/tmpdir/", "alpine:3.18.0", "rm", "-rf", "/tmpdir/project")
```
agoncharov-reef, line 78: Hm, exactly for this reason some people don't add current user to `docker` group, so the command won't work without `sudo`, and running the whole nox session with superuser rights (`sudo nox -s lint`) looks not good to me. I don't know a good solution to this, do you?

CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Getting started
+
+Main goal of the project is to provide a quick way to create production ready Django apps.
+Due to complexity of such goal solution implemented is highly opinionated and may not fit everyone's needs.
+
+- all changes must comply with [Reef Technologies conventions](https://github.com/reef-technologies/handbook#73-conventions)
+- before putting a lot of effort into a PR with a feature, please open an issue to discuss it first to avoid wasting time on something that may not be merged.
+  - the rule of thumb is that if it is not useful for 90% of projects, it should not be included
+
```
agoncharov-reef, line 13: Love it

noxfile.py
```diff
@@ -0,0 +1,110 @@
...
+    session.run('ruff', 'check', '--fix', '.')
+    run_readable(session, mode="fmt")
+
+
+@nox.session(python=PYTHON_DEFAULT_VERSION)
+def lint(session):
+    """Run linters in readonly mode."""
+    session.run('pip', 'install', '-e', '.[lint]')
+    session.run('ruff', 'check', '--diff', '.')
+    run_readable(session, mode="check")
```
agoncharov-reef, line 58: Is there any sense of running checks on non-crufted project? Cause it will cause errors where template tags are present, for example:
```
error: Failed to parse {{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tasks.py:1:2: unexpected token '%'
error: Failed to parse {{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py:1:2: unexpected token '%'
```

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/email.py
```diff
@@ -14,7 +14,7 @@
 MIMEType = TypeVar('MIMEType')
 
 
-@lru_cache(maxsize=None)
+@cache
```
agoncharov-reef, line 17: Now that I see it, I think we should limit it :) Wdyt?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tests/conftest.py
```diff
@@ -1,4 +1,4 @@
-from typing import Generator
+from collections.abc import Generator
```
agoncharov-reef, line 1: Didn't know it's the right way to do it! :-O

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -20,25 +19,37 @@
...
 
+def run_readable(session, mode="fmt"):
+    session.run(
+        'docker',
+        'run',
+        '--rm',
+        '-v', f'{ROOT.absolute()}:/data',
+        '-w', '/data',
+        '-u', f'{os.geteuid()}:{os.getegid()}',
+        'ghcr.io/bobheadxi/readable:v0.5.0@sha256:423c133e7e9ca0ac20b0ab298bd5dbfa3df09b515b34cbfbbe8944310cc8d9c9',
```
agoncharov-reef, line 30: deja-vu

noxfile.py
mjurbanski-reef: The docker guide and our own Handbook encourages you to have docker accessible from your main user so I'm not too worried that somebody is going to do `sudo root` here - if someone did custom setup then I is up to them to work out new problems they created for themselves.

As for the problem that is here - i.e. filesystem ownership issue and potential solutin - well, there is podman... but I don't think we should be introducing it here as podman compose is not as advanced as docker compose. Additionally if we start spawning container with different uid - without a proper migration of pre-existing files (i.e. old redis/postgres databases) its going to break on startup with permission issues.

tl;dr Such a change needs more thought put into it, which is why I proposed a hacky solution for the sake of CI only right now. It has 0 impact on production.

noxfile.py
mjurbanski-reef: it doesn't which is why I excluded it in config:
https://github.com/reef-technologies/cookiecutter-rt-django/pull/114/files/85372c943bdccac053616486772f7527b022a963#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R30-R32

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tests/conftest.py
mjurbanski-reef: neither did I, but it made sense when I seen that https://peps.python.org/pep-0585/#implementation covers more than list/dict/set/...

{{cookiecutter.repostory_name}}/noxfile.py
mjurbanski-reef: I don't believe there is anything we can do about this, but I'm open to suggestions

noxfile.py
agoncharov-reef: Ok, since we came nowhere on slack, let's leave it as you proposed

noxfile.py
mjurbanski-reef: I though about using `@cache` here was that I need `session` param to create the project. 
I implemented another approach, which technically doesn't use globals, but it is somewhat hackish. I would use proper decorator class if not that it IMO made code less readable.

agoncharov-reef (approved):

PR #124, opened 2023-08-01
switch to latest CPython (dropping python 3.9 compatiblity)

mjurbanski-reef:


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "cookiecutter-rt-django"
 description = "RT Django project template - don't install try to install this as package, use it as cruft template"
 version = "0"
 license = {file = "LICENSE"}
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
mjurbanski-reef, line 6: Im on purpose not changing this in pyproject.toml of crufted project since that would motivate `ruff` to updated everything to `python3.11+` syntax :
a) making next `cruft update` much bigger in terms of diff size
b) making it very hard to go back to 3.9 if for some weird reason you would like to

agoncharov-reef:
Have you ensured that all requirements support it?

mjurbanski-reef:
> Have you ensured that all requirements support it?

the singular test is passing (so at very least they install&import properly), and we use it on other project as well, so yea, kinda?
I did not verify the optional ones.

agoncharov-reef (approved):
LEEEESGO!!!

PR #125, opened 2023-08-08
Update deps & limit dependabot to security updates only

mjurbanski-reef:
so dependabot was bothering me with multiple PRs updating the same things; this PR will update most of them & limit dependabot to security updates only

### DIFF
--- a/{{cookiecutter.repostory_name}}/.github/dependabot.yml
+++ b/{{cookiecutter.repostory_name}}/.github/dependabot.yml
@@ -4,7 +4,9 @@ updates:
     directory: "/app/src"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 0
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repostory_name}}/.github/dependabot.yml
```diff
@@ -4,7 +4,9 @@ updates:
     directory: "/app/src"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 0
```
mjurbanski-reef, line 7: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file

jsuchan-reef (approved):
lgtm

PR #133, opened 2023-09-09
replace CSP_REPORT_URL with CSP_REPORT_TO

mpnowacki-reef:
I'm not sure CSP_REPORT_URL was ever a thing


cookiecutter.json
```diff
@@ -13,7 +13,7 @@
     "sentry_dsn": "",
     "csp_enabled": "n",
     "csp_report_only": "y",
-    "csp_report_url": "{% if cookiecutter.csp_enabled == 'y' %}https://sentry.io/api/-1/security/?sentry_key=key{% endif %}",
+    "CSP_REPORT_TO": "{% if cookiecutter.csp_enabled == 'y' %}https://sentry.io/api/-1/security/?sentry_key=key{% endif %}",
```
mjurbanski-reef, line 16: this seems accidental

mjurbanski-reef (changes requested):
https://django-csp.readthedocs.io/en/latest/configuration.html
while _URL wasn't, apparently CSP_REPORT_URI was a thing 

it also seems deprecated https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri but also recommended to set as report_to may not be supported by all browsers...


and most importantly if I am reading this correctly REPORT_TO shouldn't be an URL - so the currently proposed change doesn't fix anything

mpnowacki-reef:
@mjurbanski-reef you're right. I thought it worked because warnings about bad CSP stopped appearing in my browser lol :) Anyway, I did the fix because I thought it was causing firebase google log in to not work, turns out it was sth different, so let's shelve this for the moment

PR #135, opened 2023-09-17
Make app use host dir instead of volume for multiproc data storage and exporting

agoncharov-reef:


mjurbanski-reef (approved):
AFAIK it would be better to keep the named volume instead local mount (due performance, easier cleanup, less permission issues), but apparently we do local volume for redis and postgres as well so this is at very consistent.

PR #137, opened 2023-09-20
fix nginx not properly resolving new IPs after redeployment

mjurbanski-reef:
there could be a better way to fix it, but at the moment this is most cost-optimal way. 

I would argue nginx-restart wasn't that costly in the first place.


jsuchan-reef (approved):
if this fixes a problem then let's do it

PR #140, opened 2023-10-20
Add promtail

mlech-reef:


{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -185,7 +185,53 @@ services:
...
+        positions:
+          filename: /tmp/positions.yaml
+        clients:
+          - url: "${LOKI_URL}/api/prom/push"
+            basic_auth:
+              username: "${LOKI_USER}"
+              password: "${LOKI_PASSWORD}"
+            external_labels:
+              client: "${LOKI_CLIENT}"
+              client_server_group: "${LOKI_CLIENT_SERVER_GROUP}"
```
mlech-reef, line 219: We don't need client and client_server_group from the security point of view or even from the identification point of view, but I added them just to be compatible with label-based multitenancy that we have in prometheus

cookiecutter.json
```diff
@@ -6,6 +6,11 @@
     "monitoring": "y",
     "monitor_view_execution_time_in_djagno": "y",
     "monitor_tcpstat": "y",
+    "log_aggregating": "y",
+    "log_aggregator_url": "https://loki.reef.pl",
+    "log_aggregator_user": "acme",
+    "log_aggregator_client": "{{ cookiecutter.log_aggregator_user }}",
+    "log_aggregator_client_server_group": "{{ cookiecutter.log_aggregator_client }}_{{ cookiecutter.repostory_name }}",
```
mjurbanski-reef, line 13: lets call these `monitoring_client*`
since these should be common for all monitoring tools

cookiecutter.json
```diff
@@ -6,6 +6,11 @@
     "monitoring": "y",
     "monitor_view_execution_time_in_djagno": "y",
     "monitor_tcpstat": "y",
+    "log_aggregating": "y",
+    "log_aggregator_url": "https://loki.reef.pl",
+    "log_aggregator_user": "acme",
```
mjurbanski-reef, line 11: acme has some meaning? IMO best if we leave USER out of cookiecutter templating vars as only use it sees is another template and it HAS to be changed

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -185,7 +185,53 @@ services:
...
+              - host: unix:///var/run/docker.sock
+                refresh_interval: 5s
+            relabel_configs:
+              - source_labels: ["__meta_docker_container_name"]
+                regex: "/(.*)"
+                target_label: "container"
+              - source_labels: ["__meta_docker_container_log_stream"]
+                target_label: "logstream"
+        EOF
+        /usr/bin/promtail -config.file=/etc/promtail/config.yml
```
mjurbanski-reef, line 232: since we are already mounting single files, lets mount entrypoint script as well; 
1) docker-compose won't be that big :P
2) we can edit it properly with help of IDE

mjurbanski-reef (approved):
:+1:

cookiecutter.json
mlech-reef: But it is not for monitoring (Prometheus), it is only for log aggregation tool (Loki). If change the name, it would be misleading

cookiecutter.json
mlech-reef: I don't understand. It must be properly set by user. "Acme" is just a dummy value (popular word for fake company), the same as "project" for repo name is something that must be changed.

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
mlech-reef: I agree, but :D so I can either crate/delete it during setup-xxx.sh or do similar to nginx - create `promtail` directory in root dir and use it only for prod.

mjurbanski-reef: `app/envs/prod/promtail-entrypoint.sh` sounds good

cookiecutter.json
mjurbanski-reef: doesn't sound like it bring any value here, so lets throw it out from here and have just empty `LOKI_USER=` in `.env.template`, just like `LOKI_PASSWORD=` and any other prod-only credential

cookiecutter.json
mjurbanski-reef: It kinda has to be same as in Prometheus tho or we will go insane, so IDK about it being misleading.

It would be also nice to use this value in sentry_sdk; all in effort to be able to somewhat correlate events across these different monitoring systems. But do as you will.

cookiecutter.json
mlech-reef: Fine for me, but I need to remove `log_aggregator_client` and `log_aggregator_client_server_group` as well

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
mlech-reef: `envs/prod/promtail.yml`

cookiecutter.json
mlech-reef: Doesn't matter. It has to be removed anyway as we will not have cookiecutter.log_aggregator_user

mpnowacki-reef:
I see this is approved, any reason for not merging it now?

mlech-reef:
> I see this is approved, any reason for not merging it now?

Have been waiting for README update first

PR #142, opened 2023-10-24
increase db connection reliability when working with pgbouncer

mjurbanski-reef:
to be consider:
disabling DB-side cursors; they should be disabled with transaction based pgbouncer pooling anyhow and they will break on reconnection

this fixed OperationError that happens sometimes with pgbouncer (especially if it has some proxy in front of it).
Fixed issues when working with Vultr postgresql db.


mjurbanski-reef:
this not safe to use by default due to https://github.com/jdelic/django-dbconn-retry/issues/41

PR #143, opened 2023-11-08
allow for local running of spellchecker

mjurbanski-reef:
Right now it is pretty annoying that I only get notified about some silly typo (or false positive) only after submitting PR. 
The configuration was also problematic in a command line...
As for running it under `lint` CI job - it is slightly cheaper to do, since github actions bills by the execution time, but rounds it up to full minutes - so this is in fact improvement in that way as well :)

### DIFF
--- a/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
@@ -8,18 +8,6 @@ env:
   PYTHON_DEFAULT_VERSION: "3.11"
 
 jobs:
-{%- if cookiecutter.ci_use_spellchecker == "y" %}
-  spellchecker:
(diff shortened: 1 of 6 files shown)


{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
```diff
@@ -1,25 +1,15 @@
 name: Run linter and tests on pull requests
```
agoncharov-reef, line 1: "on push and pull requests"

agoncharov-reef (approved):

{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
mjurbanski-reef: "" :D

PR #145, opened 2023-11-20
add shellcheck

mjurbanski-reef:
I have ignored a significant amount of check here, mostly to reduce time required to implement this in a way that should not break anything.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -134,6 +134,7 @@ def crufted_project(session):
         session.run("cruft", "create", ".", "--output-dir", tmpdir.name, "--no-input")
         project_path = Path(tmpdir.name) / "project"
         with session.chdir(project_path):
+            session.run("git", "init")
             session.run('./setup-dev.sh')
     else:
         project_path = Path(tmpdir.name) / "project"
--- a/noxfile.py
+++ b/noxfile.py
@@ -37,7 +37,8 @@ def run_readable(session, mode="fmt"):
         '-v', f'{ROOT.absolute()}:/data',
         '-w', '/data',
         'ghcr.io/bobheadxi/readable:v0.5.0@sha256:423c133e7e9ca0ac20b0ab298bd5dbfa3df09b515b34cbfbbe8944310cc8d9c9',
-        mode, *MD_PATHS
+        mode, *MD_PATHS,
+        external=True,
     )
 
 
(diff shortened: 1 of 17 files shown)


noxfile.py
```diff
@@ -28,7 +30,28 @@
...
+        cmd_result = subprocess.run(cmd, check=True, text=True, capture_output=True)
+        file_list.extend(cmd_result.stdout.splitlines())
+    file_paths = [Path(p) for p in file_list]
+    return file_paths
+
+
+def list_files(suffix: str | None = None) -> list[Path]:
+    """List all non-files not-ignored by git."""
+    file_paths = _list_files()
+    if suffix is not None:
```
jsuchan-reef, line 49: how about passing the `suffix` into `_list_files()` and deal with it there, since you're already iterating through the list there?

{{cookiecutter.repostory_name}}/bin/backup-db.sh
```diff
@@ -19,7 +19,7 @@ else
   $DUMP_DB_TO_STDOUT > "$TARGET"
 
   if [ -n "${EMAIL_HOST:-}" ] && [ -n "${EMAIL_TARGET:-}" ]; then
-    ${SCRIPT_DIR}/backup-db-to-email.sh "${TARGET}"
+    "${SCRIPT_DIR}"/backup-db-to-email.sh "${TARGET}"
```
jsuchan-reef, line 22: is this what the shellcheck recommended? I have never seen quotes like this, usually it's the whole path in quotes instead of just the variable. Did it give any explanation why is it recommended this way? 🤔

jsuchan-reef (approved):

noxfile.py
mjurbanski-reef: that wouldn't allow to leverage the cache if start calling this multiple times for different kinds of files

for example, it would be nice to use it in `readable` formatter, since its wildcard support is also subpar, but I opted to skip that in this PR just to cut some corners and save time.

{{cookiecutter.repostory_name}}/bin/backup-db.sh
mjurbanski-reef: non explanation was given, they are AFAIK practicall identical and typically, I also prefer `"${SCRIPT_DIR}/backup-db-to-email.sh"` form, but what I prefer even more is not carring about this detail :D

PR #147, opened 2023-11-24
fix db backup breaking production db session when used with pg_bouncer

mjurbanski-reef:


{{cookiecutter.repostory_name}}/bin/restore-db.sh
```diff
@@ -7,6 +7,6 @@ target="$1"
 
 DOCKER_NETWORK=$(get_db_docker_network)
 
-zcat "$target" | docker run -i --rm --network "$DOCKER_NETWORK" postgres:14.0-alpine psql "$DATABASE_URL"
+zcat "$target" | docker run -i --rm --network "$DOCKER_NETWORK" postgres:16-alpine psql "$DATABASE_URL"
```
c0ntribut0r, line 10: Docker-compose uses v14:
```
  db:
    image: postgres:14.0-alpine
```
I believe that backup/restore scripts should have same version as the database itself.

c0ntribut0r (approved):

PR #149, opened 2023-12-05
properly secure /business-metrics endpoint

mjurbanski-reef:


{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
```diff
@@ -160,6 +164,14 @@ server {
         proxy_pass http://app:8000/metrics;
     }
 
+    location /business-metrics {
```
jsuchan-reef, line 167: I guess it's not a real issue here, but `location` without a trailing slash can lead to path traversing ( random google result: https://rioasmara.com/2021/12/27/path-traversal-via-nginx-alias/ )

jsuchan-reef (approved):

jsuchan-reef:
> properly secure

securing against the well-known path traversal due to missing trailing slash, or something else?

mjurbanski-reef:
@jsuchan-reef 

1) it blocks requests to /business-metrics endpoint from people accessing site by it is proper domain (as checked on nginx level)
2) it allows access to /business-metrics endpoint only to clients verified by our cert - we use the same auth mechanism for all prometheus endpoint 

1&2 are done by two separate virtual host configs in nginx which

PR #155, opened 2024-02-22
Use pdm

vbaltrusaitis-reef:


{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -17,11 +19,15 @@
...
 
-# In CI, use Python interpreter provided by GitHub Actions
-if CI:
-    nox.options.force_venv_backend = 'none'
+def install(session: nox.Session, *args):
+    groups = []
+    for group in args:
+        groups.extend(['--group', group])
+    session.run('pdm', 'lock', '--check', external=True)
+    session.run('pdm', 'sync', *groups, external=True)
```
mjurbanski-reef, line 30: shouldn't this be just
```suggestion
    session.run('pdm', 'install', '--check', *groups, external=True)
```
?

{{cookiecutter.repostory_name}}/pyproject.toml
```diff
@@ -1,20 +1,64 @@
 [project]
 name = "{{cookiecutter.repostory_name}}"
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
mjurbanski-reef, line 3:
```suggestion
requires-python = "==3.11"
```
for library it would make sense to support a range, but for application it makes package resolution needlessly complicated

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -111,7 +117,8 @@ def format_(session):
 @nox.session(python=PYTHON_DEFAULT_VERSION)
 def lint(session):
     """Run linters in readonly mode."""
-    session.run('pip', 'install', '-e', '.[lint]')
+    # Check if lockfile is synchronized with pyproject.toml
```
mjurbanski-reef, line 120: seems like this check is done always in install(...)
```suggestion
```

mjurbanski-reef (approved):
gj; minor notes added

{{cookiecutter.repostory_name}}/noxfile.py
vbaltrusaitis-reef: Thanks. I wrote this comment when I was doing the check here instead of `install` function, forgot to delete.

{{cookiecutter.repostory_name}}/pyproject.toml
```diff
@@ -1,20 +1,64 @@
...
+    {% endif -%}
+    {% if cookiecutter.use_fingerprinting == "y" -%}
+    "django-fingerprint-rt @ git+https://github.com/reef-technologies/django-fingerprint-rt.git@220b659",
+    {% endif -%}
+]
 
 [project.optional-dependencies]
-format = ["ruff"]
+gunicorn = [
+    "gunicorn==20.1.0",
```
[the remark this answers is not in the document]
vbaltrusaitis-reef: Agreed. I was thinking the same, just didn't want to attract wrath from whoever might care about that micro-optimization (gunicorn was initially installed in docker only). :D

PR #157, opened 2024-02-26
Fix pdm version in CI

vbaltrusaitis-reef:


{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           python-version: {% raw %}${{ env.PYTHON_DEFAULT_VERSION }}{% endraw %}
           cache: "pip"
       - name: Install dependencies
-        run: python -m pip install --upgrade nox pdm
+        run: python -m pip install --upgrade nox 'pdm>=2.12,<3'
```
emnoor-reef, line 26: Looking at [pdm changelog](https://github.com/pdm-project/pdm/blob/main/CHANGELOG.md) of v2.11.0 and v2.12.0, it looks like they introduce breaking changes in minor releases. Though we don't a lot of pdm commands/flags.

emnoor-reef (approved):

{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
vbaltrusaitis-reef: Yeah, I'd say it's not that crucial to never encounter breaking pdm changes because that would most likely result in an error in CI, not some silent corruption in the installed packages. I'm not sure which exact changes you meant but a couple examples that I saw did only result in warnings for using deprecated stuff, not errors. And I do see mentions of "fixing a freaking change" in changelog, it doesn't seem that they're breaking stuff willy-nilly in every minor release.

PR #160, opened 2024-03-02
Fix docker compose and ecr name

mpnowacki-reef:


{{cookiecutter.repostory_name}}/devops/tf/main/files/docker-compose.yml
```diff
@@ -45,7 +44,7 @@ services:
     logging: &exporter_logging
       driver: journald
       options:
-        tag: '{% raw %}{{.Name}}{% endraw %}'
+        tag: '{% raw %}{###{.Name}###}{% endraw %}'
```
mjurbanski-reef, line 47: question: the previous value seemed to be correct; why was it changed?

mjurbanski-reef (approved):
lgtm; left one question since there is one thing I don't know how will it work

{{cookiecutter.repostory_name}}/devops/tf/main/files/docker-compose.yml
mpnowacki-reef: the previous value never worked, it broke on `terraform apply`

PR #161, opened 2024-03-04
Make the CI show format diff when format check fails

vbaltrusaitis-reef:
`--diff` does the same as `--check`, except it also shows the relevant diff.

### DIFF
--- a/{{cookiecutter.repostory_name}}/noxfile.py
+++ b/{{cookiecutter.repostory_name}}/noxfile.py
@@ -126,7 +126,7 @@ def lint(session):
     session.run("codespell", ".")
     run_shellcheck(session, mode="check")
     run_readable(session, mode="check")
-    session.run("ruff", "format", "--check", ".")
+    session.run("ruff", "format", "--diff", ".")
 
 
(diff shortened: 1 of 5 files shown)


{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -126,7 +126,7 @@ def lint(session):
     session.run("codespell", ".")
     run_shellcheck(session, mode="check")
     run_readable(session, mode="check")
-    session.run("ruff", "format", "--check", ".")
+    session.run("ruff", "format", "--diff", ".")
```
mjurbanski-reef, line 129: please add this line to https://github.com/reef-technologies/cookiecutter-rt-django/blob/master/noxfile.py#L121 as well

mjurbanski-reef (approved):

{{cookiecutter.repostory_name}}/noxfile.py
vbaltrusaitis-reef: Added that, formatted, also added `.git-blame-ignore-revs`.

PR #167, opened 2024-03-26
Silence invalid Host header warnings

jakub-cichalewski-reef:
[NICE-155](https://reeftechnologies.atlassian.net/browse/NICE-155) [RT-143](https://reeftechnologies.atlassian.net/browse/RT-143)

Silence `django.security.DisallowedHost` logger in sentry

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/settings.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/settings.py
@@ -302,7 +302,7 @@ if SENTRY_DSN := env("SENTRY_DSN", default=""):
     import sentry_sdk
     from sentry_sdk.integrations.celery import CeleryIntegration
     from sentry_sdk.integrations.django import DjangoIntegration
-    from sentry_sdk.integrations.logging import LoggingIntegration
+    from sentry_sdk.integrations.logging import LoggingIntegration, ignore_logger
     from sentry_sdk.integrations.redis import RedisIntegration
 
     sentry_sdk.init(  # type: ignore
@@ -318,3 +318,4 @@ if SENTRY_DSN := env("SENTRY_DSN", default=""):
             ),
         ],
     )
+    ignore_logger("django.security.DisallowedHost")


{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
```diff
@@ -98,6 +98,11 @@ server {
             return 425;
         }
         {%- endif %}
+
+        {% if cookiecutter.nginx_filter_bad_host_headers == 'y' -%}
+        include /etc/nginx/config_helpers/ignore_bad_host.conf;
+        {%- endif %}
+
```
mjurbanski-reef, line 105: seeing how we have
>     server_name               ${NGINX_HOST} {% if cookiecutter.nginx_redirect_www_prefix == 'n' %}www.${NGINX_HOST}{% endif %};

on top, can you elaborate how does this help?

{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
```diff
@@ -132,6 +137,10 @@ server {
     access_log                /dev/stdout;
     error_log                 /dev/stderr info;
 
+    {% if cookiecutter.nginx_filter_bad_host_headers == 'y' -%}
+    include /etc/nginx/config_helpers/ignore_bad_host_allow_ip.conf;
+    {%- endif %}
```
mjurbanski-reef, line 142: monitoring by us is typically done through autodiscovery from AWS or over service provider
I.e. we query by IPs/domains assigned to a particular ec2 machine, i.e. it is almost never $NGINX_HOST

so this this won't really help us.

Either it should allow all valid domains/adresses here OR we just silence Invalid HTTP_HOST header on sentry level.

mjurbanski-reef (changes requested):

{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
jakub-cichalewski-reef: This is a mistake, I put it in the wrong block

{{cookiecutter.repostory_name}}/nginx/templates/default.conf.template
jakub-cichalewski-reef: Nginx would be more performant, but it seems like resolving domains could be more complexity than it's worth. I changed to ignoring the logger on sentry level

PR #173, opened 2024-04-10
docker compose nginx healthcheck for docker compose deployments

mjurbanski-reef:


{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -123,7 +123,14 @@ services:
     image: 'ghcr.io/reef-technologies/nginx-rt:v1.2.1'
     restart: unless-stopped
     healthcheck:
-      test: wget -q --spider 0.0.0.0:80 || exit 1
+      test: [
```
mjurbanski-reef, line 126: probably could have done with wget, but curl is present in that image and lighter

The core errors was that default nginx vhost returned 444, i.e. NO response, which always failed, so the actual fix is asking for  `$NGINX_HOST` vhost.

The rest is just some additional polishing I added while looking at it.

PR #174, opened 2024-04-10
Healthcheck endpoint

adal-chiriliuc-reef:
Added `/healthcheck` endpoint.

It's enabled if `cookiecutter.monitoring` is set. Should I always enable it, or create a new `cookiecutter.healthcheck` env?

Checks database and redis. Returns 200 or 500 status code with status json:

```jsonc
// HTTP 200
{
  "database_ok": true,
  "redis_ok": true,
  "all_ok": true
}
```
```jsonc
// HTTP 500
{
  "database_ok": true,
  "redis_ok": false,
  "redis_error": "ConnectionError('Error 111 connecting to localhost:8379. Connection refused.')",
  "all_ok": false
}
```
```jsonc
// HTTP 500
{
  "database_ok": false,
  "database_error": "ProgrammingError('relation \"healthcheck_model\" does not exist\\nLINE 1: DELETE FROM \"healthcheck_model\"\\n                    ^\\n')",
  "redis_ok": true,
  "all_ok": false
}

```

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
@@ -6,6 +6,7 @@ from fingerprint.views import FingerprintView
 {% endif -%}
 {% if cookiecutter.monitoring == "y" %}
 from .{{cookiecutter.django_default_app_name}}.business_metrics import metrics_manager
+from .{{cookiecutter.django_default_app_name}}.healthcheck import healthcheck_view
 from .{{cookiecutter.django_default_app_name}}.metrics import metrics_view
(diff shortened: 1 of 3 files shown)


mjurbanski-reef:
why not https://github.com/revsys/django-health-check ?

adal-chiriliuc-reef:
I looked at that lib, even took some inspiration from it.

Seemed a bit heavy, I will check it again in more detail.

mjurbanski-reef:
@adal-chiriliuc-reef unless there is some strong argument against it, let's not reinvent the wheel here and use what seems to be quite popular solution for this problem.  That solution would be in line with our tenets https://github.com/reef-technologies/handbook/tree/master?tab=readme-ov-file#what-does-high-performance-mean .

PR #176, opened 2024-04-15
Add structlog support

agoncharov-reef:
Structured logging is great. Easier to read logs, easier to debug.

We talked too much about it. Maybe it's time to just add it, and if we don't like it we can `git revert`.


mjurbanski-reef (approved):
love it, just please make sure CI also loves it before merging :D

agoncharov-reef:
Before CI added / after CI added
drugs-man-1_1834017i

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tasks.py
```diff
@@ -1,12 +1,14 @@
 {%- if cookiecutter.use_celery == 'y' -%}
 from celery.utils.log import get_task_logger
+import structlog
```
mjurbanski-reef, line 3: :P

not sure I assumed ruff will fail CI here, so but I guess it is not enabled by default so it doesnt

PR #178, opened 2024-04-17
Add websocket support with django-channels

agoncharov-reef:
This PR adds django-channels as well as default consumer which can handle incoming messages.


{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/consumers.py
```diff
@@ -0,0 +1,45 @@
...
+        log.info("connected", scope=self.scope)
+
+    async def disconnect(self, code: int | str) -> None:
+        log.info("disconnected", scope=self.scope, code=code)
+
+    async def receive(self, text_data: str | None = None, bytes_data: bytes | None = None) -> None:
+        with bound_contextvars(text_data=text_data, bytes_data=bytes_data):
+            log.debug("message received")
+
+            adapter = TypeAdapter(Annotated[Union[*self.MESSAGE_HANDLERS.keys()], Field(discriminator="type")])
```
mjurbanski-reef, line 27: can we create one adapter and reuse it, instead of recreating it on every message received? seems wasteful

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/consumers.py
```diff
@@ -0,0 +1,45 @@
...
+            try:
+                message: BaseModel = adapter.validate_json(text_data)
+            except ValidationError as exc:
+                errors = exc.errors()
+                log.debug("message parsing failed", errors=errors)
+                return
```
mjurbanski-reef, line 33: we need error handler called here

mjurbanski-reef (approved):
seems great; added some notes

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/consumers.py
agoncharov-reef: I'm dumb, what do you mean? I left this place as "put your own logic to react to errors here"

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/consumers.py
agoncharov-reef: You never leave me a chance to be lazy :D

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/consumers.py
mjurbanski-reef, line 31: I would expect something like
`self.send(text_data=Error(message=str(exc).json())`
to let the client know what happend

Then again, this is asynchronus, so perhaps this also requires that each message also has identifier of the requests so you can know for requests this is a reply for?

agoncharov-reef: Yep exactly, and the identifier is project/task-specific. May be UUID, or int, or string. So I left it for developer to decide how to implement responses.

PR #179, opened 2024-04-21
Added healthcheck endpoint

adalc-reef:
Uses https://github.com/revsys/django-health-check

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/settings.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/settings.py
@@ -261,7 +261,9 @@ CHANNEL_LAYERS = {
 {%- endif %}
 
 {%- if cookiecutter.monitoring == "y" %}
-REDIS_URL = env("REDIS_URL")
+REDIS_HOST = env("REDIS_HOST")
+REDIS_PORT = env.int("REDIS_PORT")
+REDIS_URL = f"redis://{REDIS_HOST}:{REDIS_PORT}"
 {%- endif %}
 
 {%- if cookiecutter.use_celery == "y" %}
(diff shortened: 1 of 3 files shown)


mjurbanski-reef:
@adal-chiriliuc-reef please remove `disable celery healthchecks` check

It just proves it works correctly if it fails when celery is not working.
I see no downside to this behavior - its not like any automatic process is watching it locally. Tell me if you see something here I dont.
At the same time we do want it by default in prod when celery is enabled.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/settings.py
```diff
@@ -249,6 +260,10 @@ def wrapped(*args, **kwargs):
 }
 {%- endif %}
 
+{%- if cookiecutter.monitoring == "y" %}
+REDIS_URL = env("REDIS_URL", default="")
```
mjurbanski-reef, line 264:
```suggestion
REDIS_URL = env("REDIS_URL")
```

mjurbanski-reef: actually, this is silly

why aren't we using REDIS_HOST and REDIS_PORT ?

mjurbanski-reef: that wasn't the forbidden part, please change it to:
```
REDIS_HOST = env("REDIS_HOST")
REDIS_PORT = env.int("REDIS_PORT")
REDIS_URL = f"redis://{REDIS_HOST}:{REDIS_PORT}"
```

sorry for not being clear in the beginning

mjurbanski-reef: thank you and sorry again for confusion

mjurbanski-reef (approved):

mjurbanski-reef:
no, it just waits for you to merge it :P

PR #182, opened 2024-05-09
Remove pdm workaround

adalc-reef:
Remove PDM workaround, problem is caused by incompatibility between `pdm` and the handbook recommended `virtualenvwrapper` setup in `.bashrc`, but we are removing that - https://github.com/reef-technologies/handbook/pull/122


mjurbanski-reef:
there are still some people (2 at the moment) using it, so I guess lets keep it for now?

PR #185, opened 2024-05-29
RT-147 allauth

kkowalski-reef:
Intorduces optional integration with django-allauth
(!) new cookiecutter variables: 
- `use_allauth`
- `allauth_trust_external_emails`
- `use_allauth_*`

Streamlined config available for providers:
- apple
- atlassian
- discord
- facebook
- github
- gitlab
- google
- microsoft
- openid_connect (any generic OIDC provider)
- twitter

### DIFF
--- a/{{cookiecutter.repostory_name}}/README.md
+++ b/{{cookiecutter.repostory_name}}/README.md
@@ -4,8 +4,6 @@
 
 - - -
 
-Skeleton of this project was generated with `cookiecutter-rt-django`, which sometimes gets upgrades that are easy to retrofit into already older projects.
-
 # Base requirements
 
 - docker with [compose plugin](https://docs.docker.com/compose/install/linux/)
@@ -16,7 +14,7 @@ Skeleton of this project was generated with `cookiecutter-rt-django`, which some
 # Setup development environment
 
 ```sh
-$ ./setup-dev.sh
+./setup-dev.sh
 docker compose up -d
 cd app/src
 pdm run manage.py wait_for_database --timeout 10
@@ -25,6 +23,8 @@ pdm run manage.py runserver
 ```
 
 # Setup production environment (git deployment)
+
+<details>
 
 This sets up "deployment by pushing to git storage on remote", so that:
 
@@ -100,6 +100,9 @@ If one wants to deploy other branch, force may be used to push desired branch to
 ```sh
 git push --force production local-branch-to-deploy:master
 ```
+
+</details>
+
 {% if cookiecutter.use_allauth == 'y' %}
 # External auth (OAuth, OpenID connect etc.)
 To configure an external authentication mechanism, usually you must acquire a "client ID" and a "client secret".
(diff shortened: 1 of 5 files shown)


agoncharov-reef (commented):
I remember allauth has some basic templates which one should always override with custom ones, maybe some custom templates should already be included into cookiecutter? I'm not 100% sure tho.

Also, it allows registration of two types as far as i remember: with sending confirmation email and without it, but here it's not covered at all.

Ideally I would like to have cookiecutter questions be like
- enable registration?
- require email confirmation?
- enable google login?
- enable apple ID login?
- ...

So "adding allauth support" is not about adding it to requirements, `INSTALLED_APPS` and including instructions in README but about answering few questions and get everything set up for you. With current PR I still have to do stuff manually. 

I don't know what original task was so I have no clue whether what I wrote is valid or not, just my thoughts.

mjurbanski-reef:
>require email confirmation?

I cannot really imagine a case in which you allow autoregistration and do not require email confirmation.
It should be either confirmed by SSO (that is an email provider as well (this is important to verify)) or by email directy.

cookiecutter.json
```diff
@@ -13,6 +13,18 @@
...
+    "use_allauth_apple": "n",
+    "use_allauth_atlassian": "n",
+    "use_allauth_discord": "n",
+    "use_allauth_facebook": "n",
+    "use_allauth_github": "n",
+    "use_allauth_gitlab": "n",
+    "use_allauth_google": "n",
+    "use_allauth_microsoft": "n",
+    "use_allauth_openid_connect": "n",
+    "use_allauth_twitter": "n",
```
0xE111, line 27: Nitpick (not a call-to-action):
I may be the only person who suggests what I gonna suggest, but I'd rather make it a list, cause we already have too many cookiecutter questions:
```
# cookiectutter
"allauth_providers": "apple,atlassian,google,...",

# other files
{% if 'google' in cookiecutter.allauth_providers %}
```

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
```diff
@@ -16,7 +16,9 @@
 
 urlpatterns = [
     path("admin/", site.urls),
+    {%- if cookiecutter.use_allauth != "y" %}
```
0xE111, line 19: I'd use if-else, so we see alternatives in the same place:
```
{% if use_allauth == 'y' %}
     path('accounts/', include('allauth.urls')),
{% else %}
    path("", include("django.contrib.auth.urls")),
{% endif %}
```

{{cookiecutter.repostory_name}}/envs/dev/.env.template
```diff
@@ -70,3 +70,50 @@ BACKUP_B2_BUCKET=
...
 BACKUP_B2_KEY_SECRET=
 BACKUP_LOCAL_ROTATE_KEEP_LAST=
+
+{% if cookiecutter.use_allauth == "y" -%}
+{%- if cookiecutter.use_allauth_apple == "y" -%}
+APPLE_LOGIN_CLIENT_ID=
+APPLE_LOGIN_SECRET=
+APPLE_LOGIN_KEY=
+APPLE_LOGIN_CERTIFICATE_PRIVATE_KEY=
+{% endif %}
```
0xE111, line 80: Shouldn't these changes also go to terraform env templates?

0xE111 (approved):

{{cookiecutter.repostory_name}}/README.md
```diff
@@ -100,6 +100,45 @@ If one wants to deploy other branch, force may be used to push desired branch to
 ```sh
 git push --force production local-branch-to-deploy:master
 ```
+{% if cookiecutter.use_allauth == 'y' %}
+# External auth (OAuth, OpenID connect etc.)
```
mjurbanski-reef, line 104: please make this section collapsible; see  https://github.com/reef-technologies/cookiecutter-rt-django/pull/191/files#diff-378103025370115268a1f51cd31dc9694912f0739345121e4d6f3c473097b7d6R27  for example

PR #187, opened 2024-06-09
Refresh migration metrics every metrics export

### LINKED ISSUE #186: Metrics for migrations do not update after migration
The django_migrations_unapplied_total metric is not resetting the values even after migrations are completed, causing the alert never to go off. (unless the app is restarted)

Also, the founder of Prometheus node-exporter raised the same issue here: 
https://github.com/korfuri/django-prometheus/issues/405

Reproduce:
Rolled back some migrations using:
`pdm run manage.py migrate auth 0005`

Then, I started the server:
`pdm run manage.py runserver`

Observed the metrics at /metrics endpoint:
```
django_migrations_unapplied_total{connection="default"} 7.0 
django_migrations_applied_total{connection="default"} 15.0
```

Ran the migrate command on a different process:
`pdm run manage.py migrate`
Rechecked the metrics at /metrics endpoint. Observed no update in the metrics.

Restarted the server:
`pdm run manage.py runserver`

Observed the updated metrics at /metrics endpoint:
```
django_migrations_unapplied_total{connection="default"} 0.0
django_migrations_applied_total{connection="default"} 22.0
```

Related ticket:
https://reeftechnologies.atlassian.net/jira/software/c/projects/NICE/boards/41?selectedIssue=NICE-325

luca-medeiros-reef:
Closes #186 

Every metrics call will get the latest migration-related metrics.

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
@@ -5,6 +5,7 @@ import os
 import prometheus_client
 from django.http import HttpResponse
 from django_prometheus.exports import ExportToDjangoView
+from django_prometheus.migrations import ExportMigrations
 from prometheus_client import multiprocess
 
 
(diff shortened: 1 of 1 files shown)


mjurbanski-reef (commented):
to me this seems good (assuming `ExportMigrations()` doesn't take too much time (>=100ms) - and it seems it doesn't )

but please try to solve this issue in upstream as well, so we have a longterm solution

luca-medeiros-reef:
Yeah, I was worried it would add too much load (as it has to query the db too).
A quick single-user locust run shows relevant load but is still in an acceptable range for a Prometheus with a 5~30s refresh rate.
Old: avg 34ms
New: avg 82ms
image

luca-medeiros-reef:
We could further improve and refresh db migration metrics less often

luca-medeiros-reef:
I'll check what can be done upstream. Given that the issue was raised over a year ago with no response, I doubt they will go further to merge a PR regarding this. (Wonder if this is expected behavior..)

Will look into Prometheus custom-collectors and come up with a PR upstream.

PR #195, opened 2024-07-24
Add dead letter queue support

agoncharov-reef:
nuff said

### DIFF
--- a/{{cookiecutter.repostory_name}}/README.md
+++ b/{{cookiecutter.repostory_name}}/README.md
@@ -145,6 +145,28 @@ If an SSO provider supports the OIDC protocol, it can be set up as a generic OID
 3. A "profile" page is available at `/accounts/`
 
 {% endif %}
+
+{% if cookiecutter.use_celery == 'y' %}
+# Background tasks with Celery
+
+## Dead letter queue
+
+There is a special queue named `dead_letter` that is used to store tasks
+that failed for some reason.
+
+A task should be annotated with `on_failure=send_to_dead_letter_queue`.
+Once the reason of tasks failure is fixed, the task can be re-processed
+by moving tasks from dead letter queue to the main one ("celery"):
+
+    manage.py move_tasks "dead_letter" "celery"
+
+If tasks fails again, it will be put back to dead letter queue.
+
+To flush add tasks in specific queue, use
+
+    manage.py flush_tasks "dead_letter"
+
+{% endif %}
 {% if cookiecutter.monitoring == 'y' %}
 # Monitoring
 
--- a/{{cookiecutter.repostory_name}}/README.md
+++ b/{{cookiecutter.repostory_name}}/README.md
@@ -151,6 +151,7 @@ If an SSO provider supports the OIDC protocol, it can be set up as a generic OID
 
 ## Dead letter queue
 
+<details>
 There is a special queue named `dead_letter` that is used to store tasks
 that failed for some reason.
 
@@ -165,6 +166,7 @@ If tasks fails again, it will be put back to dead letter queue.
 To flush add tasks in specific queue, use
 
     manage.py flush_tasks "dead_letter"
+</details>
 
 {% endif %}
 {% if cookiecutter.monitoring == 'y' %}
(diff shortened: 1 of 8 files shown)


{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
```diff
@@ -33,4 +37,14 @@ def metrics_view(request):
         )
     else:
         return ExportToDjangoView(request)
+
+
+num_tasks_in_queue = {}
+for queue in settings.CELERY_TASK_QUEUES:
+    gauge = prometheus_client.Gauge(
```
ppolewicz, line 44: ❤️

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tasks.py
```diff
@@ -1,13 +1,40 @@
...
+    Once the reason of tasks failure is fixed, the task can be re-processed
+    by running a worker on dead letter queue:
+
+        celery -A mrc.celery worker -Q dead_letter
+
+    If tasks fails again, it will be put back to dead letter queue. This can
+    potentially create an infinite loop, so make sure to fix the reason of
+    failure before re-processing the task, or stop the worker after a few
+    retries. The process of re-processing tasks is supposed to be manual,
+    thus there are no complicated mechanisms to prevent infinite loops.
```
ppolewicz, line 25: Shouldn't we add a dead dead letter queue for this and then the fixing processor would be configured to pick stuff up from dead_letter and run it but on failure put in dead_dead_letter, then if this happens you can fix the code in such a way that it's not sending anything to dead_dead, run through the whole dead and finally run like a special task which takes everything from dead_dead and moves to dead or runs on dead_dead and puts to dead? I doubt we'll always be perfect and there isn't anything that would warn us about the infinite loop, so putting in a process to handle it properly seems practical to me. It's a good moment to think about how to handle it.

Also this sounds like something that should be described somewhere in the documentation of our framework, not just in this docstring here.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tasks.py
```diff
@@ -1,13 +1,40 @@
...
+    """
+    if task.app.conf.task_always_eager:
+        return  # do not run failed task again in eager mode
+
+    logger.debug(
```
ppolewicz, line 30:
```suggestion
    logger.warning(
```

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/tasks.py
agoncharov-reef: I just added a command to move tasks from one queue to another. No infinite loops involved.

https://github.com/reef-technologies/cookiecutter-rt-django/commit/322fc5e89692b00a158da25ee3b9402ff9edacde

{{cookiecutter.repostory_name}}/README.md
```diff
@@ -144,6 +144,28 @@ If an SSO provider supports the OIDC protocol, it can be set up as a generic OID
 2. Allauth "social users" are just an extension to regular django users. When someone logs in via allauth, a django user model will also be created for them.
 3. A "profile" page is available at `/accounts/`
 
+{% endif %}
+
+{% if cookiecutter.use_celery == 'y' %}
+# Background tasks with Celery
```
mjurbanski-reef, line 150: please add `<details>` tags so this is collapsible

PR #201, opened 2024-10-24
fix: double empty lines when use_channels == "n" causes lint error

zyzniewski-reef:


agoncharov-reef:
Fixed by 24f127390dfc1692a20da81b3d9feac06de57e84

PR #203, opened 2024-10-29
Expose celery metrics

agoncharov-reef:
I tried both 
1) standalone dockerized celery exporter - https://github.com/danihodovic/celery-exporter
2) celery-flower

The former has more metrics but requires separate image and does not have a label by task (as far as i remember), so i switched back to flower. 

However, i also needed a queue length metric, so I added it manually into our django app (`CustomCeleryCollector`).

Grafana is already configured to show the metrics.
Screenshot_2024-10-29_14-15-39

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
@@ -19,7 +19,7 @@ class RecursiveMultiProcessCollector(multiprocess.MultiProcessCollector):
         return self.merge(files, accumulate=True)
 
 
-if (is_multiprocess := bool(os.environ.get("PROMETHEUS_MULTIPROC_DIR"))):
+if is_multiprocess := bool(os.environ.get("PROMETHEUS_MULTIPROC_DIR")):
     registry = prometheus_client.CollectorRegistry()
     RecursiveMultiProcessCollector(registry)
 else:
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
@@ -1,13 +1,15 @@
 {%- if cookiecutter.monitoring == "y" -%}
 import glob
 import os
-from functools import partial
+from collections.abc import Iterator
 
 import prometheus_client
 from django.conf import settings
 from django.http import HttpResponse
 from django_prometheus.exports import ExportToDjangoView
 from prometheus_client import multiprocess
+from prometheus_client.core import REGISTRY, GaugeMetricFamily, Metric
+from prometheus_client.registry import Collector
 
 from ..celery import get_num_tasks_in_queue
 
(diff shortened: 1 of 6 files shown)


{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
```diff
@@ -28,6 +30,7 @@ def metrics_view(request):
     if os.environ.get(ENV_VAR_NAME):
         registry = prometheus_client.CollectorRegistry()
         RecursiveMultiProcessCollector(registry)
+        registry.register(CustomCeleryCollector())
```
olzhasar-reef, line 33: Do we need to do it on each request, it looks like we already registered the thing on the line 54?

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
```diff
@@ -36,12 +39,17 @@ def metrics_view(request):
...
-    num_tasks_in_queue[queue.name] = gauge
-    gauge.set_function(partial(get_num_tasks_in_queue, queue.name))
+class CustomCeleryCollector(Collector):
+    def collect(self) -> Iterator[Metric]:
+        num_tasks_in_queue = GaugeMetricFamily(
+            "celery_queue_len",
+            "How many tasks are there in a queue",
+            labels=("queue",),
+        )
+        for queue in settings.CELERY_TASK_QUEUES:
```
olzhasar-reef, line 49: The same thing here, I don't think we need to iterate queues on each collect() invocation

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -109,11 +109,12 @@ services:
     env_file: ./.env
     environment:
       - DEBUG=off
+      - FLOWER_TASK_RUNTIME_METRIC_BUCKETS=1,2,3,5,10,20,30,45,60,120,180,240,300,600,inf
     command: celery --app={{cookiecutter.django_project_name}} --broker="${CELERY_BROKER_URL}" flower --basic_auth="${CELERY_FLOWER_USER}:${CELERY_FLOWER_PASSWORD}"
     depends_on:
       - celery-worker
     ports:
-      - 5555:5555
+      - 127.0.0.1:5555:5555
```
olzhasar-reef, line 117: Why do we need to expose a port here in the first place?

olzhasar-reef (changes requested):

{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
agoncharov-reef: Celery flower has basic http auth (so it's quite dumb) or oauth (so it's too clever for me). So not really handy to expose it to the internet. But we want to access flower. So we have `cadm <server> tunnel-flower` command which does SSH tunneling from server to `localhost:5555`. However, since flower on server is dockerized, ssh won't see it unless we expose the port to host.

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/{{cookiecutter.django_default_app_name}}/metrics.py
agoncharov-reef: This actually seems OK to me, what's wrong? On each metrics collection we get list of queues and for each queue we ask redis queue length.

olzhasar-reef: `settings.CELERY_TASK_QUEUES` is a static value that won't be changing during the app runtime, so is there a reason to iterate and create a new metric each time? I expect prometheus to be smart enough to overwrite it, but why do it in the first place?

PR #207, opened 2024-11-28
Set up vulnrelay (vulnerability scanning)

olzhasar-reef:


{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
```diff
@@ -0,0 +1,5 @@
+ENV=prod
+DD_URL=
```
mpnowacki-reef, line 2: let's add the default url here

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
```diff
@@ -0,0 +1,5 @@
+ENV=prod
```
mpnowacki-reef, line 1: this is only used by the dsn, right? what's the idea here? to use the same dsn as the project? if so then the ENV should default to `vulnrelay-prod`.

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
```diff
@@ -0,0 +1,5 @@
...
+DD_URL=
+DD_API_KEY=
+DD_PRODUCT=
```
mpnowacki-reef, line 4: this is not explained, there should be a mention of the convention: product=client_server_group

mpnowacki-reef (changes requested):
posted some questions

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: The idea was to use the same sentry project for all servers, but use the environment field to differentiate between them.
And the environment is being constructed from DD_PRODUCT and ENV:
```
    environment = f"{settings.DD_PRODUCT} - {settings.ENV}"
```

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: I don't think exposing urls of our infrastructure services in the open source repository is a good idea

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: That's a good point, but I was thinking this should be somewhere in our internal handbook in the relevant section about vulnerabilities management. Here we just referring to whatever was created in DefectDojo. Wdyt?

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
mpnowacki-reef: okay maybe let's leave a hint where to look

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
mpnowacki-reef: using the same dsn for all project is a security risk. I'd rather have them separate, even if separated from the project's DSN

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
mpnowacki-reef: I'm not sure, leaving it in another place is gonna be gelpful, people will just forget about it

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: I don't get it, how exactly are multiple sentry DSNs safer than a single one?
When it comes to various bugs that may arise in production, having all of them in a single place seems much more convenient to manage. Also, creating a separate sentry setup for each new project we ship can be daunting.

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: Added some clarifying comments to the `.env` file

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
mpnowacki-reef: the security risk is: if one DSN gets compromised a malicious actor can spawn errors for all nodes, and cycling the DSN would require changing configs of all deployments. If creating projects is troubling then it is a problem we can solve with a steady supply of evaluation candidates - sentry API allows for creation of projects. It could be a part of CADM or whatever.

{{cookiecutter.repostory_name}}/envs/prod/.vuln.env.template
olzhasar-reef: Added a link to the internal handbook in the README_AWS

PR #209, opened 2024-12-26
Migrate to uv

olzhasar-reef:


{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -8,25 +8,31 @@
...
 
 def install(session: nox.Session, *args):
     groups = []
-    for group in args:
-        groups.extend(["--group", group])
-    session.run("pdm", "install", "--check", *groups, external=True)
+    for arg in args:
+        groups.extend(["--group", arg])
+
+    uv_env = getattr(session.virtualenv, "location", os.getenv("VIRTUAL_ENV"))
```
mpnowacki-reef, line 28: what's the use case for virtualenv not being set in the session object but being given by an env var?

mpnowacki-reef (approved):

PR #211, opened 2025-01-17
Don't import celery in monitoring if it's disabled

agoncharov-reef:


agoncharov-reef:
Fixed by [804dcd1](https://github.com/reef-technologies/cookiecutter-rt-django/commit/804dcd1357657302597b630f053a02ae3b1308a9)

PR #213, opened 2025-02-04
Switch CI workflow from pdm to uv

agoncharov-reef:
I recently crufted a new project from latest cookiecutter and got a bunch of errors during CI process. Turns out we switched to uv for dependency management but haven't updated CI scripts, which is addressed by this PR.

I removed every mention of PDM except build backend in `pyproject.toml`, since [uv doesn't have its own build backend yet](https://github.com/astral-sh/uv/issues/8779).

Tested it on brand new crufted repo, all CI steps (lint + test) pass.

### DIFF
--- a/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
@@ -23,9 +23,9 @@ jobs:
         with:
           python-version: {% raw %}${{ env.PYTHON_DEFAULT_VERSION }}{% endraw %}
       - name: Install uv
-        uses: astral-sh/setup-uv@v4
+        uses: astral-sh/setup-uv@v5
         with:
-          version: "0.5.11"
+          version: "0.5.x"
       - name: Install nox
         run: python -m pip install --upgrade nox
       - name: Run linters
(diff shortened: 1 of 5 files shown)


{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
```diff
@@ -21,9 +22,12 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: {% raw %}${{ env.PYTHON_DEFAULT_VERSION }}{% endraw %}
-          cache: "pip"
-      - name: Install dependencies
-        run: python -m pip install --upgrade nox 'pdm>=2.12,<3'
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
```
olzhasar-reef, line 26: Why not setup-uv@v5? :)

{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
```diff
@@ -38,15 +42,18 @@ jobs:
...
+        with:
+          version: "0.5.11"
```
olzhasar-reef, line 48: We could pin a minor version instead:
https://github.com/astral-sh/setup-uv?tab=readme-ov-file#install-a-version-by-supplying-a-semver-range

olzhasar-reef (approved):

PR #216, opened 2025-02-24
Refactor backups & add backups monitoring

agoncharov-reef:
Total refactoring of backups.

- Instead of setting up backups manually on host, the `backups` container is automatically started with everything set up within. One deploys an app, he gets backups working automatically.
- The container does two things: runs cron and serves backup metrics on port `8000`
- We check whether latest backup is not currupted by running `pg_restore -l` on it and searching for `TABLE DATA public django_migrations` in the output. It's dumb but better than nothing for now. For backups in b2, we download only first 10MiB of data, and run the above command on that chunk - so that we don't download the whole large backup. It seems to be enough to read the ToC of postgres archive (I checked manually).
- Backups are now auto-created during deployment (on poor man's deployment only though)
- Tested poor man's setup on Timas; haven't tested AWS setup
- Metrics are already set up in prometheus (`with_backups=True`) and Grafana:
Screenshot_2025-02-24_18-53-43


{{cookiecutter.repostory_name}}/backups/bin/serve_metrics.py
```diff
@@ -0,0 +1,201 @@
...
+
+    backup_count.set(num_backups)
+
+    if first_backup:
+        first_backup_time.set(first_backup.created_at.timestamp())
+
+    if last_backup:
+        last_backup_time.set(last_backup.created_at.timestamp())
+        log.debug("Checking newest backup", last_backup=last_backup)
+        is_operational = manager.check_is_operational(last_backup)
```
emnoor-reef, line 178: opinion: checking if a backup is operational every time we update metrics is a bit too much I think, specially when using b2.

{{cookiecutter.repostory_name}}/backups/bin/rotate-local-backups.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+if [ -n "${BACKUP_LOCAL_ROTATE_KEEP_LAST:-}" ]; then
+    echo "Rotating backup files - keeping ${BACKUP_LOCAL_ROTATE_KEEP_LAST} last ones"
+    files_to_delete=$(find /var/backups -name "*.dump.zstd" | sort -r | tail -n "+${BACKUP_LOCAL_ROTATE_KEEP_LAST}")
+    echo "$files_to_delete" | xargs rm
```
emnoor-reef, line 5:
```suggestion
    echo "$files_to_delete" | xargs --no-run-if-empty rm
```

emnoor-reef (approved):

olzhasar-reef:
Why not a separate package for this though?
We have a lot of projects that are way behind the current cookiecutter state, and we seem to consistently increase the upgrade cost by introducing large changes to the template.

agoncharov-reef:
@olzhasar-reef It's a docker container, idk how to make a package from it.
I know that there is this problem of syncing old stuff with latest cookiecutter, but I think benefits are bigger than drawbacks, i.e. we are ready to pay the price of upgrading old projects.
Also, moving it to separate package doesn't mean that upgrade costs go away - now you'll need to upgrade that package in old projects, and track compatibility.
You may open a discussion on s3 and invite me, if we all decide some better design then i'll be able to implement it quickly now that everything is still fresh in my head.

olzhasar-reef:
@agoncharov-reef

> It's a docker container, idk how to make a package from it.

It's as easy as creating a separate repository with the functionality and a Dockerfile and configuring CI to push it to a public registry. The same way as we do with [nginx](https://github.com/reef-technologies/docker-nginx-rt) for example. And the template will just use the image. We could even use the latest tag with watchtower for automatic updates if the latter is a concern.

> Also, moving it to separate package doesn't mean that upgrade costs go away - now you'll need to upgrade that package in old projects, and track compatibility.

Applying template changes where an additional container has been added to the docker-compose file with some configuration is a no brainer. But reviewing changes associated with essentially a standalone application is a completely different level of mental overhead. And we will have to deal with that overhead in each project that we want to upgrade. There might be completely different people responsible for the upgrades of each repo and they'd have to immerse themselves into this context each time.

Speaking of sociocracy, I honestly don't even see the point of a separate discussion with the whole team for these kinds of questions. I am convinced that everyone would just agree that cookiecutter upgrades are a huge problem and if there's any way to make them less expensive, then it should be utilized.

PR #218, opened 2025-02-27
Configure gcr mirrors to avoid possible dockerhub limits

olzhasar-reef:
- **Add mirror.gcr.io to cloud-init and packer deployments**
- **Use images from mirror.gcr.io in /dev/docker-compose.yml**

### DIFF
--- a/{{cookiecutter.repostory_name}}/bin/prepare-os.sh
+++ b/{{cookiecutter.repostory_name}}/bin/prepare-os.sh
@@ -53,6 +53,16 @@ if [ ! -x "${DOCKER_BIN}" ] || [ ! -x "${DOCKER_COMPOSE_INSTALLED}" ]; then
     usermod -aG docker "$USER"
 fi
 
+if [ ! -x "${JQ_BIN}" ]; then
+  apt-get -y install jq
+fi
+
+if [ ! -f /etc/docker/daemon.json ]; then
+  echo '{ "registry-mirrors": ["https://mirror.gcr.io"] }' > /etc/docker/daemon.json
+else
+  jq '.["registry-mirrors"] += ["https://mirror.gcr.io"]' /etc/docker/daemon.json > /etc/docker/daemon.tmp && mv /etc/docker/daemon.tmp /etc/docker/daemon.json
+fi
+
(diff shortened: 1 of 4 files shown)


olzhasar-reef:
I tested the `cloud-init.yml` changes with [multipass](https://canonical.com/multipass).
The `/etc/docker/daemon.json` is being created as expected. However, the docker itself somehow fails to install.
I tried installing it manually in the VM, but it says:
```
ubuntu@test-rt:~$ sudo apt-get install docker-ce                                                                                                                            
Reading package lists... Done                                                                                                                                               
Building dependency tree                                                                                                                                                    
Reading state information... Done                                                                                                                                           
Package docker-ce is not available, but is referred to by another package.                                                                                                  
This may mean that the package is missing, has been obsoleted, or                                                                                                           
is only available from another source                                                                                                                                       
However the following packages replace it:                                                                                                                                  
  docker-ce-cli:amd64  
```
I assume it's because I'm on an ARM machine and the VM has the same architecture instead of the original amd64.
But anyways, if it worked before, there shouldn't be any reasons for it to break now.

{{cookiecutter.repostory_name}}/bin/prepare-os.sh
```diff
@@ -8,6 +8,7 @@ SENTRY_CLI="$(command -v sentry-cli || true)"
 B2_CLI="$(command -v b2 || true)"
 AWS_CLI="$(command -v aws || true)"
 JQ_BIN="$(command -v jq || true)"
+USER="$(id -un 1000)"
```
emnoor-reef, line 11: I never thought about using id to get username. It will fail for insecure hosts (where only `root` user exists, no user id 1000), but fuck them.

Previously I used `${SUDO_USER:-$USER}` in a script that needed to get the username of the user that ran the script with sudo.

emnoor-reef (approved):

{{cookiecutter.repostory_name}}/bin/prepare-os.sh
olzhasar-reef: I believe I learned that trick while configuring my home server once :)

PR #219, opened 2025-02-27
Fix promtail config

olzhasar-reef:
- **Set tenant_id in promtail config**
- **Update docker-compose to persist promtail positions in a volume**

### DIFF
--- a/{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
+++ b/{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
@@ -216,6 +216,7 @@ services:
     volumes:
       - ./promtail:/etc/promtail
       - /var/run/docker.sock:/var/run/docker.sock:ro
+      - promtail-data:/var/lib/promtail
     command:
       - -config.file=/etc/promtail/config.yml
       - -config.expand-env=true
@@ -249,3 +250,4 @@ volumes:
   backend-static:
   gunicorn-socket:
   nodeexporter_collectors:
+  promtail-data:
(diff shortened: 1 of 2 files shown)


{{cookiecutter.repostory_name}}/envs/prod/docker-compose.yml
```diff
@@ -216,6 +216,7 @@ services:
     volumes:
       - ./promtail:/etc/promtail
       - /var/run/docker.sock:/var/run/docker.sock:ro
+      - promtail-data:/var/lib/promtail
```
olzhasar-reef, line 219: Without persisting promtail positions, it will try to upload the entire log history on restart, which is not a very nice thing to do

{{cookiecutter.repostory_name}}/promtail/config.yml
```diff
@@ -3,10 +3,11 @@ server:
   grpc_listen_port: 0
 
 positions:
-  filename: /tmp/positions.yaml
+  filename: /var/lib/promtail/positions.yaml
 
 clients:
   - url: "${LOKI_URL}/loki/api/v1/push"
+    tenant_id: "${LOKI_TENANT_ID:=rt}"
```
olzhasar-reef, line 10: This is likely not needed in our current setup, but it doesn't break anything and will make it easier if we'd like to use the [multi-tenancy](https://grafana.com/docs/loki/latest/operations/multi-tenancy/) feature in future.

emnoor-reef (approved):

PR #220, opened 2025-03-26
Fix unix socket healthcheck + add alive endpoint

olzhasar-reef:
- **Add missing host header to unix domain socket healthcheck**
- **Remove backups container from dev docker-compose**
- **Introduce dedicated /alive/ endpoint for health checking**

### DIFF
--- a/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
+++ b/{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.admin.sites import site
+from django.http import HttpResponse
 from django.urls import include, path{% if cookiecutter.use_rest_framework == "y" %}, re_path{% endif %}
 
 {% if cookiecutter.use_rest_framework == "y" %}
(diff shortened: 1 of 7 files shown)


{{cookiecutter.repostory_name}}/envs/dev/docker-compose.yml
```diff
@@ -23,31 +23,3 @@ services:
       - ./db/data:/var/lib/postgresql/data
     ports:
       - ${POSTGRES_PORT}:5432
-
-  backups:
```
agoncharov-reef, line 27: The backups container is needed if you want to... well... restore backups on local machine. Seems like a frequent use-case during development :)

{{cookiecutter.repostory_name}}/app/src/{{cookiecutter.django_project_name}}/urls.py
```diff
@@ -14,7 +15,13 @@
 from .{{cookiecutter.django_default_app_name}}.metrics import metrics_view
 {%- endif %}
 
+
+def alive_view(request):
+    return HttpResponse(b"ok")
+
+
 urlpatterns = [
+    path("alive/", alive_view),
```
agoncharov-reef, line 24: I'd just inline `path("ping/", lambda _: HttpResponse()),`, wdyt?

agoncharov-reef (commented):
> I am not sure if we even use that health-check app somewhere.

I also think that the health check app is doing too much:

```python
    "health_check",
    "health_check.db",
    "health_check.cache",
    "health_check.storage",
    "health_check.contrib.migrations",
    "health_check.contrib.psutil",
    "health_check.contrib.redis",
    "health_check.contrib.celery",
    "health_check.contrib.celery_ping",
```

And I don't see it used by anything, maybe it's a good to ask if anyone has complains about it being deleted, and delete it completely? Cause it's confusing rn.

{{cookiecutter.repostory_name}}/envs/dev/docker-compose.yml
olzhasar-reef: Having a long-running container for that seems a bit excessive, I think there should be a way to restore backups by spinning the container on demand.
But restored for now

olzhasar-reef:
@agoncharov-reef I also question the necessity of having the `healthcheck` package, but let's address that in a separate PR.

PR #236, opened 2026-01-28
Do not persist PROMETHEUS_MULTIPROC_DIR

olzhasar-reef:
Store prometheus metrics in a container-local folder instead of mounting
a host directory.
The $PROMETHEUS_MULTIPROC_DIR is a runtime directory for storing metrics
accumulated for a current process. It is not meant to be persisted
long-term. Prometheus itself is engineered with outages in mind and
storing .db for already dead processes does not bring any value (but
does lead to infinitely growing folders on productions servers).


kacper-wolkiewicz-reef (approved):
LGTM

agoncharov-reef:
https://github.com/reef-technologies/cookiecutter-rt-django/pull/259

PR #260, opened 2026-03-29
Conventional commits

agoncharov-reef:
- Introduced "conventional commits" procedure
- Added nox session to check for some commit rules
    - conventional commits naming
    - squashed branches
    - no merge commits
- Added publishing to slack on new tag

### DIFF
--- a/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
         with:
           # checkout pull request HEAD commit instead of merge commit
           ref: {% raw %}${{ github.event.pull_request.head.sha }}{% endraw %}
+
           fetch-depth: 0
           fetch-tags: true
       - uses: ./.github/actions/setup-uv-nox
--- a/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
+++ b/{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
@@ -10,43 +10,51 @@ env:
   ENV_FILL_MISSING_VALUES: 1
 
 jobs:
+  conventional_commits:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # checkout pull request HEAD commit instead of merge commit
+          ref: {% raw %}${{ github.event.pull_request.head.sha }}{% endraw %}
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: ./.github/actions/setup-uv-nox
+      - name: Check following commits convention
+        run: uvx nox -vs check_commits
 {% if cookiecutter.ci_use_linter %}
   linter:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Install uv
-        uses: astral-sh/setup-uv@v7
-        with:
-          version: "0.10.x"
-      - name: Install nox
-        run: uv tool install --with pyyaml nox
+      - uses: ./.github/actions/setup-uv-nox
       - name: Run linters
         run: uvx nox -vs lint
+      - name: Check for missing migrations
+        run: uvx nox -vs check_missing_migrations
 {% endif %}
   test:
     timeout-minutes: 10
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-      - name: Install uv
-        uses: astral-sh/setup-uv@v7
-        with:
-          version: "0.10.x"
+      - uses: ./.github/actions/setup-uv-nox
       - name: Prepare environment
         run: ./setup-dev.sh
       - name: Run dockerized services
         run: docker compose up -d --wait
       - name: Run migrations
         run: cd app/src && uv run python manage.py wait_for_database --timeout 120 && uv run python manage.py migrate
-      - name: Install nox
-        run: uv tool install --with pyyaml nox
       - name: Run unit tests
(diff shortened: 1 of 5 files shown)


{{cookiecutter.repostory_name}}/README.md
```diff
@@ -20,6 +20,24 @@ uv run manage.py migrate
...
 ```
 
+# Commits and releases (Commitizen)
+
+[Commitizen](https://commitizen-tools.github.io) may be used to create commits and releases in a standardized way.
+It is not required for creating commits and releases, but it is handy for bumping versions and generating changelogs.
+
+```sh
+# install the tool once
+uv tool install commitizen==4.13.9
```
kkowalski-reef, line 30: imo no need to pin cz version

{{cookiecutter.repostory_name}}/README.md
```diff
@@ -20,6 +20,24 @@ uv run manage.py migrate
...
+
+# create a conventional commit
+# https://commitizen-tools.github.io/commitizen/commands/commit/
+cz commit
+
+# make a release (bump version, update changelog, create a tag)
+# https://commitizen-tools.github.io/commitizen/commands/bump/
+cz bump && uv lock
```
kkowalski-reef, line 38: If `cz bump` creates the commit, the lockfile will not land in that commit.

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -162,3 +175,103 @@ def test(session):
             "{{cookiecutter.django_project_name}}",
             *session.posargs,
         )
+
+
+VERSION_TAG_PATTERN = "v[0-9]+.[0-9]+.[0-9]+"
```
kkowalski-reef, line 180: May be good enough to start, but:

https://packaging.python.org/en/latest/specifications/version-specifiers/#appendix-parsing-version-strings-with-regular-expressions

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -162,3 +175,103 @@ def test(session):
...
+    for i, line in enumerate(lines):
+        accumulated_length += len(line)
+        if accumulated_length > limit:
+            return "".join(lines[:i]) + "\n… (truncated)\n"
+
+    assert False
+
+
+@nox.session(python=PYTHON_VERSION)
+def check_commits(session: nox.Session):
```
kkowalski-reef, line 220: Not sure about this bundle in general. I would rather have it split into separate sessions and move the check that shouldn't run on master to a separate workflow, but:

1. "No merge commits" check - we rarely have merge commits on branches themselves, the check doesn't prevent **merging** to master - only merging nonlinear branches, it failing on master requires a force push on master to fix or pushing another tag to silence it. Makes more sense to enforce via github repo settings on the templated project.
2. Squash check - we are still allowed to do multiple commits per PR, we have an agreement to squash WIPs so they don't pollute the git history post-merge. IMO not enforceable via CI.

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -162,3 +175,103 @@ def test(session):
...
+        if num_commits > 1:
+            session.error(f"Branch '{current_branch}' has {num_commits} commits since diverging from '{base_branch}'. Please squash your commits before merging.")
+
+
+@nox.session(python=PYTHON_VERSION)
+def publish_changelog(session: nox.Session):
+    import requests
+
+    if not _get_current_tag(session):
+        return
```
kkowalski-reef, line 254: Would be cleaner to have a "release" workflow on tags or something similar. The "ci" workflow is getting bloated and these conditions make it even harder to reason about it - this runs on all pushes to master and PRs, but only really does something if there is a version tag on the commit.

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -162,3 +175,103 @@ def test(session):
...
+    current_branch = _get_current_branch(session)
+    if current_branch and current_branch != base_branch:
+        num_commits = int(session.run("git", "rev-list", "--count", f"{base_branch}..HEAD", silent=True, external=True).strip())
+        if num_commits > 1:
+            session.error(f"Branch '{current_branch}' has {num_commits} commits since diverging from '{base_branch}'. Please squash your commits before merging.")
+
+
+@nox.session(python=PYTHON_VERSION)
+def publish_changelog(session: nox.Session):
+    import requests
```
kkowalski-reef, line 251: Isn't requests installed only after `install(session, "commits_check")`?

{{cookiecutter.repostory_name}}/noxfile.py
```diff
@@ -146,6 +147,18 @@ def type_check(session):
         session.run("mypy", "--config-file", "mypy.ini", ".", *session.posargs)
 
 
+@nox.session(python=PYTHON_VERSION)
+def check_missing_migrations(session):
+    install(session, "check_missing_migrations")
```
kkowalski-reef, line 152: I like the check but i don't see this dependency group here?

{{cookiecutter.repostory_name}}/.github/workflows/ci.yml
```diff
@@ -10,43 +10,52 @@ env:
...
       - name: Run unit tests
         run: uvx nox -vs test
       - name: Stop dockerized services
         if: success() || failure()
         run: docker compose down -v
+  publish_changelog:
+    needs: [conventional_commits, test]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
kkowalski-reef, line 58: Looks like this needs tags fetched?

kkowalski-reef (changes requested):

PR #277, opened 2026-06-23
feat: add flags to toggle aws, direct and vultr deploy generation

kacper-wolkiewicz-reef:
Add use_aws_deploy, use_direct_deploy and use_vultr_deploy flags (all default true) so each deploy type can be excluded from a generated project. Whole deploy directory trees are removed in post_gen_project.sh because cookiecutter cannot drop a populated directory via a conditional name. Shared files (setup-prod.sh, deploy.sh, prepare-os.sh) are kept while any deploy that uses them is still enabled.


cookiecutter.json
```diff
@@ -25,6 +25,9 @@
     "csp_enabled": false,
     "nginx_compression_enabled": false,
     "nginx_tls_early_data_enabled": false,
+    "use_aws_deploy": true,
+    "use_direct_deploy": true,
+    "use_vultr_deploy": true,
```
agoncharov-reef: They seem to be mutually exclusive? If yes (i think so) then I'd rather have a single choice (https://cookiecutter.readthedocs.io/en/stable/advanced/choice_variables.html)
Then I'd use something like this on standalone files:
- `{% if cookiecutter.deployment == "vultr" %}filename.py{% endif %}`
- `{% if cookiecutter.deployment in ["vultr", "aws"] %}filename.py{% endif %}`

hooks/post_gen_project.sh
```diff
@@ -1,5 +1,26 @@
...
+
+{% if not cookiecutter.use_direct_deploy %}
+rm -f letsencrypt_setup.sh
+{% endif %}
+
+{% if not cookiecutter.use_direct_deploy and not cookiecutter.use_vultr_deploy %}
+rm -f deploy.sh bin/prepare-os.sh
+{% endif %}
+
+{% if not cookiecutter.use_aws_deploy and not cookiecutter.use_direct_deploy and not cookiecutter.use_vultr_deploy %}
```
agoncharov-reef: Hm, is this condition possible?

hooks/post_gen_project.sh
```diff
@@ -1,5 +1,26 @@
...
+rm -rf devops/vultr_tf devops/vultr_scripts README_vultr.md
+{% endif %}
+
+{% if not cookiecutter.use_direct_deploy %}
+rm -f letsencrypt_setup.sh
+{% endif %}
+
+{% if not cookiecutter.use_direct_deploy and not cookiecutter.use_vultr_deploy %}
+rm -f deploy.sh bin/prepare-os.sh
+{% endif %}
```
agoncharov-reef: (here and in other places)
For standalone files I'd use conditional file name. Otherwise it's easy to move/rename file and forget to do it here.

hooks/post_gen_project.sh
kacper-wolkiewicz-reef: Technically it is, you will then not get any deployment out of the box. Probably noone will use it, but better safe than sorry.

PR #282, opened 2026-07-22
Add engineering standards and Agent guidelines

jan-sauer-reef:


{{cookiecutter.repostory_name}}/AGENTS.md
```diff
@@ -0,0 +1,21 @@
...
+- the new requirements from the user or issue
+- any tests that describe the same behavior
+
+If the code, requirements, and user stories disagree, resolve the behavior deliberately instead of preserving stale documentation. Update [USER_STORIES.md](USER_STORIES.md) in the same change whenever code behavior changes.
+
+If USER_STORIES.md doesn't exist or is empty, inform your handler that they need to add user stories to ensure documentation aligns with code.
+
+## General Workflow
+
+- Prefer existing patterns used in this repository when designing and implementing new features.
```
mpnowacki-reef, line 17: it would be convenient for me to include

```
- follow the rules described in files in .local_agent_instructions. 
  they are meant for developers who need to add some clarifications, overrides regarding
  their work stations
```

{{cookiecutter.repostory_name}}/AGENTS.md
```diff
@@ -0,0 +1,21 @@
...
+- Keep documentation claims tied to code that exists in the current tree.
+- Before marking a task as done, make sure tests and linters {% if cookiecutter.ci_use_typechecker %} and type checks {% endif %} pass. The general README.md contains instructions for running these checks.
+- follow the rules described in [engineering-standards.md](engineering-standards.md)
```
mpnowacki-reef: I think maybe we should mention that each task should explicitly check for test correctness, whether they adhere to the rules in engineering_standards, sort of "heal as you go" approach.

Or maybe that should be only mentioned in a separate guidelines file, specific for the review bot, not sure.

{{cookiecutter.repostory_name}}/AGENTS.md
```diff
@@ -0,0 +1,21 @@
...
+
+If the code, requirements, and user stories disagree, resolve the behavior deliberately instead of preserving stale documentation. Update [USER_STORIES.md](USER_STORIES.md) in the same change whenever code behavior changes.
+
+If USER_STORIES.md doesn't exist or is empty, inform your handler that they need to add user stories to ensure documentation aligns with code.
+
+## General Workflow
+
+- Prefer existing patterns used in this repository when designing and implementing new features.
+- Keep documentation claims tied to code that exists in the current tree.
+- Before marking a task as done, make sure tests and linters {% if cookiecutter.ci_use_typechecker %} and type checks {% endif %} pass. The general README.md contains instructions for running these checks.
```
mpnowacki-reef, line 19: I think specific instructions on how to run tests linters etc should also be mentioned, unless they already are in the readme

mpnowacki-reef: they are, never mind

{{cookiecutter.repostory_name}}/AGENTS.md
jan-sauer-reef: Is an extra file necessary or could project maintainers just make direct edits to AGENTS.md?

jan-sauer-reef: Oh wait, this is git-ignored and for actual individual people. That makes sense. AGENTS.local.md seems to be an emerging convention for this that I've seen a few times.

{{cookiecutter.repostory_name}}/AGENTS.md
jan-sauer-reef: Doesn't hurt to have them in both. Implementing agents tend to also review their own work when the effort is set high enough.

{{cookiecutter.repostory_name}}/AGENTS.md
mpnowacki-reef: perfect!

